PR #1, opened 2021-11-17
Initial version with first utils

Jongy:
This includes extracting OOM events & exception signals (SIGSEGV, traps, ...) from kernel messages, which we plan to use in gProfiler among other components.

### DIFF
--- a/granulate_utils/linux/oom.py
+++ b/granulate_utils/linux/oom.py
@@ -5,21 +5,20 @@
 
 import re
 from collections import namedtuple
-
 from typing import Optional
 
-
-# see function in oom_kill.c:__oom_kill_process
-# example line: "Out of memory: Killed process 765074 (chrome) total-vm:38565352kB, anon-rss:209356kB, file-rss:1624kB, shmem-rss:0kB"
-# recent kernels added more fields (e.g UID, pgtables) but we don't care about them, for now; and this regex still works.
+# see function in oom_kill.c:__oom_kill_process. example string:
+# "Out of memory: Killed process 765074 (chrome) total-vm:38565352kB, anon-rss:209356kB, file-rss:1624kB, shmem-rss:0kB"
+# recent kernels added more fields (e.g UID, pgtables) but we don't care about them, for now; and this regex still
+# works.
 KILLED_PROCESS = re.compile(
-    r"(?:<\d>)?(?:\[(?P<timestamp>\d+\.\d+)\] )?(?:(?P<message>.*): )?Killed process (?P<pid>\d+) \((?P<comm>\w+)\) total-vm:(?P<total_vm>\d+)kB, anon-rss:(?P<anon_rss>\d+)kB, file-rss:(?P<file_rss>\d+)kB, shmem-rss:(?P<shmem_rss>\d+)kB"
+    r"(?:<\d>)?(?:\[(?P<timestamp>\d+\.\d+)\] )?(?:(?P<message>.*): )?Killed process (?P<pid>\d+) "
+    r"\((?P<comm>\w+)\) total-vm:(?P<total_vm>\d+)kB, anon-rss:(?P<anon_rss>\d+)kB, "
+    r"file-rss:(?P<file_rss>\d+)kB, shmem-rss:(?P<shmem_rss>\d+)kB"
 )
 KB = 1024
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+version = re.search(r'__version__\s*=\s*"(.*?)"', Path("utils/__init__.py").read_text())
+assert version is not None, "could not parse version!"
+
+with open("README.md", "r", encoding="utf-8") as fh:
+    long_description = fh.read()
+
+setuptools.setup(
+    name="utils",
```
d3dave, line 24: That's gonna conflict with https://pypi.org/project/utils/

d3dave (changes requested):

setup.py
Jongy: Wasn't planning to, at least initially. We can install directly from git.

d3dave: pip is gonna complain if someone installs that package.

Jongy: Our components run in their own container / own Python environment. So I doubt it'll affect us. We can name it `granulate-utils`.

lint.sh
```diff
@@ -0,0 +1,16 @@
...
+if [ -f venv/bin/activate ]; then
+  source venv/bin/activate
+fi
+
+check_arg=""
+if [[ "$1" = "--ci" ]]; then
+    check_arg="--check"
+fi
+
+isort --settings-file .isort.cfg .
```
d3dave, line 13: Which version of `isort` are you using? New version uses `--settings-path` and requires `-rc`

.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
+max-line-length = 120
+# options for compatibility with black - see https://black.readthedocs.io/en/stable/compatible_configs.html#flake8
+extend-ignore = E203, W503
+exclude = .git,__pycache__,build,gprofiler.egg-info,venv
```
d3dave, line 5:
```suggestion
exclude = .git,__pycache__,build,venv
```

d3dave (changes requested):

.flake8
Jongy: Hmm already had it but I didn't push
```
commit ea5b5fff91a365e30355674d5ed9edfd75837214 (HEAD -> first-utils)
Author: Yonatan Goldschmidt <yonatan.goldschmidt@granulate.io>
Date:   Wed Nov 17 02:13:19 2021 +0200

    Remove unrelated dirs

diff --git a/.flake8 b/.flake8
index 2b3aecf..4dacb92 100644
--- a/.flake8
+++ b/.flake8
@@ -2,4 +2,4 @@
 max-line-length = 120
 # options for compatibility with black - see https://black.readthedocs.io/en/stable/compatible_configs.html#flake8
 extend-ignore = E203, W503
-exclude = .git,__pycache__,build,gprofiler.egg-info,venv
+exclude = .git,__pycache__,venv
```

lint.sh
Jongy: VERSION 5.8.0.

Will update both

Jongy:
```
/home/jong/.local/lib/python3.8/site-packages/isort/main.py:1206: UserWarning: W0501: The following deprecated CLI flags were used and ignored: -rc!
  warn(
/home/jong/.local/lib/python3.8/site-packages/isort/main.py:1210: UserWarning: W0500: Please see the 5.0.0 Upgrade guide: https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0/
  warn(
```

not adding `-rc`.

d3dave: `--recursive` otherwise it doesn't recurse

Jongy: It's not a flag to 5.8.0
```
$ isort --help | grep -2 -- recursive
                        against.
  --dont-follow-links   Tells isort not to follow symlinks that are
                        encountered when running recursively.
  --filename FILENAME   Provide the filename associated with a stream.
```

setup.py
Jongy: Let's make this change if this ever bothers

Jongy:
Rebased

setup.py
Jongy: Changed in #6

d3dave (approved):

PR #2, opened 2021-11-24
Bootstrap

Jongy:


YishaiZinkin (approved):
Looks like just the usual new project boilerplate. Didn't look too much into it, in case of problems we'll just fix.
LGTM

Jongy:
Re-creating so we squash-and-merge

PR #5, opened 2021-11-24
Add proc_events.py

YishaiZinkin:


utils/linux/proc_events.py
```diff
@@ -0,0 +1,214 @@
...
+This module uses the Linux's Process Connector mechanism to invoke registered callbacks on various process events.
+To achieve this, once the first callback is registered a thread is invoked to listen to these events and execute the
+registered callbacks.
+
+This interface with the kernel isn't too well documented, but to learn more about it you can read this blog post:
+https://nick-black.com/dankwiki/index.php/The_Proc_Connector_and_Socket_Filters
+
+
+TODO: Add more callbacks.
+TODO: Use socket filter.
```
Jongy, line 12: Wdym here btw?

Jongy:
See comments from https://github.com/Granulate/gprofiler/pull/208

utils/linux/proc_events.py
```diff
@@ -0,0 +1,214 @@
...
+    @_raise_if_not_running
+    def register_exit_callback(self, callback: Callable):
+        self._exit_callbacks.append(callback)
+
+    @_raise_if_not_running
+    def unregister_exit_callback(self, callback: Callable):
+        self._exit_callbacks.remove(callback)
+
+
+_proc_events_listener = _ProcEventsListener()
```
Jongy, line 191: This `__init__` still runs some code (creates a socket, ...)

Let's push the initialization to `_ensure_thread_started` as well.

utils/linux/proc_events.py
```diff
@@ -0,0 +1,214 @@
...
+                                ]
+                            ),
+                        )
+                    )
+
+                    for callback in self._exit_callbacks:
+                        callback(event_data["pid"], event_data["tgid"], event_data["exit_code"])
+
+        # Cleanup
+        self._selector.unregister(self._socket)
```
Jongy, line 170: This needs to be in some `finally` block. If anything raises during the loop, you leak them.

I suggest that you split it into 2 functions - one that does the registration, then calls the second function in `try` and in its `finally` closes everything. The other function is just the loop

Jongy (changes requested):
.

utils/linux/proc_events.py
YishaiZinkin: Filter the messages we're getting so the thread won't get up for every event in the system

Jongy: Ah, like, a BPF filter for specific PIDs?

YishaiZinkin: Yes.
I'm somewhat concerned that this thread will introduce noticeable overhead. This filter will solve it.

Jongy: https://www.explainxkcd.com/wiki/index.php/1691:_Optimization

utils/linux/proc_events.py
```diff
@@ -206,9 +210,11 @@ def register_exit_callback(callback: Callable):
 
     The callback should receive three arguments: tid, pid and exit_code.
     """
+    _proc_events_listener: _ProcEventsListener
```
Jongy, line 219: Instead of re-defining its type, the better thing to do here IMO is to `assert _proc_events_listener is not None`. This also teaches mypy

YishaiZinkin: Cool, didn't know that

Jongy (approved):

Jongy (approved):

PR #9, opened 2021-11-26
Add basic linting in GitHub actions

Jongy:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,29 @@
+name: granulate-utils lint
+
+on: pull_request
+
+jobs:
+  lint:
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
```
YishaiZinkin, line 9: Hmmm I think we should define better what's the purpose of this repo. If it aims to serve the agents as well we should test 2.7, and if it's for the other products I don't see a reason to check 3.6 and 7 (and maybe 8?)

Jongy: This serves gProfiler and possibly other components. gProfiler supports 3.6->3.9 so we need to test them all

YishaiZinkin: Kk, we should remember to add more environments other than `ubuntu-latest` as we add logical tests

YishaiZinkin (approved):
Cool

.github/workflows/lint.yml
Jongy: This repo will deserve its own tests at some point. For now, since all of its features are used by gProfiler, I suppose that gProfiler tests cover it well enough.

PR #10, opened 2021-11-27
Start proc_events in the calling thread

Jongy:
So if an exception is raised during the bind()/_register_for_connector_events(),
it is visible in the calling thread.

### DIFF
--- a/granulate_utils/linux/proc_events.py
+++ b/granulate_utils/linux/proc_events.py
@@ -199,11 +199,18 @@ _proc_events_listener: Optional[_ProcEventsListener] = None
 def _ensure_thread_started(func: Callable):
     def wrapper(*args, **kwargs):
         global _proc_events_listener
+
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/proc_events.py
```diff
@@ -196,11 +199,18 @@ def unregister_exit_callback(self, callback: Callable):
...
+
         if _proc_events_listener is None:
-            _proc_events_listener = _ProcEventsListener()
-            _proc_events_listener.start()
+            try:
+                _proc_events_listener = _ProcEventsListener()
+                _proc_events_listener.start()
+            except Exception:
+                _proc_events_listener = None
+                raise
```
Jongy, line 209: We still leak the pipes here :/ Oh well. Next time

YishaiZinkin (approved):
LGTM

PR #11, opened 2021-11-28
Add DevKmsgProvider

d3dave:


Jongy:
> Fixing /home/runner/work/granulate-utils/granulate-utils/granulate_utils/linux/ns.py

I suppose that's my bad for merging that before the CI branch

granulate_utils/linux/devkmsg.py
```diff
@@ -0,0 +1,59 @@
...
+import errno
+import os
+import time
+from typing import List, Tuple
+
+# See linux/printk.h
+CONSOLE_EXT_LOG_MAX = 8192
+
+
+class DevKmsgProvider:
```
Jongy, line 14: What about the `Provider` interface? I suppose that can be moved into this repo as well - since we'll want to support the `EmptyProvider` for old kernels / low privileges cases. (and at some point we'll add support for kernels < 3.5 so it should also share the same interface)

Jongy:
Looks good. Push the matching gProfiler PR that uses it?

granulate_utils/linux/devkmsg.py
```diff
@@ -0,0 +1,58 @@
...
+class DevKmsgProvider(KernelMessagesProvider):
+    """
+    Provides a stream of kernel messages via /dev/kmsg. Requires Linux 3.5+.
+
+    The /dev/kmsg interface is described at Documentation/ABI/testing/dev-kmsg in the kernel source tree
+    and can be viewed at https://github.com/torvalds/linux/blob/master/Documentation/ABI/testing/dev-kmsg.
+    """
+
+    def __init__(self):
+        self.dev_kmsg_fd = os.open("/dev/kmsg", os.O_RDONLY)
```
Jongy, line 25: Open it with `open` and store the result, then use `.fileno()`. This way the result is cleaned up automatically once this object dies.

Jongy (changes requested):
.

Jongy (approved):

PR #13, opened 2021-11-29
Improve OOM regex

Jongy:
`\w+` works for e.g `java` but not for names containing `-`, which are common (`systemd-...`).

### DIFF
--- a/granulate_utils/linux/__init__.py
+++ b/granulate_utils/linux/__init__.py
@@ -12,4 +12,5 @@ def get_kernel_release() -> Tuple[int, int]:
     return int(major_str), int(minor_str)
 
 
+# TASK_COMM_LEN is 16, and it is always null-terminated, so 15.
 COMM_PATTERN = r".{0,15}"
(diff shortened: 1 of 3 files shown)


granulate_utils/linux/__init__.py
```diff
@@ -10,3 +10,6 @@ def get_kernel_release() -> Tuple[int, int]:
     """Return Linux kernel version as (major, minor) tuple."""
     major_str, minor_str = os.uname().release.split(".", maxsplit=2)[:2]
     return int(major_str), int(minor_str)
+
+
+COMM_PATTERN = r".{0,15}"
```
d3dave, line 15: not 16?

Jongy: last is always `\x00` IIRC

Jongy: I'll add a comment

d3dave: Yep, looks like it's 15 🙁

d3dave (approved):

PR #14, opened 2021-12-05
Added the common case of resolving a path from the host by using the pid=1 process

tomersa:


granulate_utils/linux/ns.py
```diff
@@ -31,6 +32,10 @@ class NsType(enum.IntFlag):
 
 libc: Optional[ctypes.CDLL] = None
 
+def resolve_host_root_links(ns_path: str) -> str:
+    assert ns_path.startswith('/'), 'Must pass full path and not relative'
```
guybortnikov, line 36:
```suggestion
    assert ns_path.startswith('/'), 'Must pass absolute path and not relative'
```

Jongy, line 36: This is asserted in the inner function, no need to check for it again.

tomersa:
Looks like the lints are failling but that's not related to these changes. 
image

Jongy:
Is this still needed? If so, please rebase, I suppose those lint issues were fixed in master already.

tomersa:
> Is this still needed? If so, please rebase, I suppose those lint issues were fixed in master already.

done

granulate_utils/linux/ns.py
```diff
@@ -63,8 +69,8 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
 
 def is_same_ns(pid: int, nstype: str, pid2: int = None) -> bool:
     return (
-        os.stat(f"/proc/{pid2 if pid2 is not None else 'self'}/ns/{nstype}").st_ino
-        == os.stat(f"/proc/{pid}/ns/{nstype}").st_ino
+            os.stat(f"/proc/{pid2 if pid2 is not None else 'self'}/ns/{nstype}").st_ino
```
Jongy, line 72: Your `black` version probably mismatches. This change is incorrect. Make sure you have the right version and run formatting again to fix it

Jongy:
You don't run local linting. `black` enforces use of `"` for strings, and you're using `'` here so I suppose that's the reason for the lint failure. Fix lint and we can merge

tomersa:
> You don't run local linting. `black` enforces use of `"` for strings, and you're using `'` here so I suppose that's the reason for the lint failure. Fix lint and we can merge

Done. Please approve

Jongy (approved):

PR #15, opened 2021-12-05
linux: ns: implement get_proc_root path both for resolved and unresolved

guybortnikov:


granulate_utils/linux/ns.py
```diff
@@ -155,3 +155,18 @@ def is_running_in_init_pid() -> bool:
...
+
+
+def get_proc_root_path(process: Process) -> str:
+    """
+    Gets /proc/<pid>/root of a given process, then a file can be read from the host mnt ns.
+    """
+    return f"/proc/{get_mnt_ns_ancestor(process)}/root"
+
+
+def get_resolved_proc_root_path(process: Process, ns_path: str) -> str:
```
d3dave, line 167: I think a better name would be `resolve_host_path`

granulate_utils/linux/ns.py
```diff
@@ -155,3 +155,18 @@ def is_running_in_init_pid() -> bool:
...
+    """
+    Gets /proc/<pid>/root of a given process, just like `get_proc_root_path` except that here we also resolve all the
+    filesystem links.
```
d3dave, line 170:
```suggestion
    Get a path in the host mount namespace pointing to path in process mount namespace.
```

d3dave (changes requested):

d3dave (approved):

PR #16, opened 2021-12-07
Added support for registering on process exec events

IlayRosenberg:
Also including a fix to the listener loop which would previously crash when handled too many events, and ignoring missing imports for the psutil library which failed `lint.sh` for me

### DIFF
--- a/granulate_utils/linux/proc_events.py
+++ b/granulate_utils/linux/proc_events.py
@@ -234,20 +234,22 @@ class _ProcEventsListener(threading.Thread):
 
 
 _proc_events_listener: Optional[_ProcEventsListener] = None
+_listener_creation_lock = threading.Lock()
 
 
 def _ensure_thread_started(func: Callable):
     def wrapper(*args, **kwargs):
         global _proc_events_listener
 
-        if _proc_events_listener is None:
-            try:
-                _proc_events_listener = _ProcEventsListener()
-                _proc_events_listener.start()
-            except Exception:
-                # TODO: We leak the pipe FDs here...
-                _proc_events_listener = None
-                raise
+        with _listener_creation_lock:
+            if _proc_events_listener is None:
+                try:
+                    _proc_events_listener = _ProcEventsListener()
+                    _proc_events_listener.start()
+                except Exception:
+                    # TODO: We leak the pipe FDs here...
+                    _proc_events_listener = None
+                    raise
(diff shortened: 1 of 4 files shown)


granulate_utils/linux/proc_events.py
```diff
@@ -116,7 +124,14 @@ def _listener_loop(self):
...
 
             for key, _ in events:
-                data = key.fileobj.recv(256)
+                try:
+                    # When stressed, reading from the socket can raise
+                    #   OSError: [Errno 105] No buffer space available
+                    # This seems to be safe to ignore, empirically no events were missed
+                    data = key.fileobj.recv(256)
+                except OSError as e:
+                    if e.errno == 105:
```
YishaiZinkin, line 133: Do you suppress other exceptions on purpose?

IlayRosenberg: woops

granulate_utils/linux/proc_events.py
```diff
@@ -116,7 +124,14 @@ def _listener_loop(self):
                 break
 
             for key, _ in events:
-                data = key.fileobj.recv(256)
+                try:
+                    # When stressed, reading from the socket can raise
+                    #   OSError: [Errno 105] No buffer space available
+                    # This seems to be safe to ignore, empirically no events were missed
```
YishaiZinkin, line 130: Interesting. Did you see it when there was a large amount of processes or on general network stress?

IlayRosenberg: Large amount of processes. I stressed it on purpose locally

YishaiZinkin: Hmm maybe I'll add packet filtering at some point which might solve this

mypy.ini
```diff
@@ -2,3 +2,6 @@
 python_version = 3.6
 warn_unused_configs = True
 exclude = venv/bin/
+
+[mypy-psutil.*]
+ignore_missing_imports = True
```
YishaiZinkin, line 7: It works for me, it seems like a local problem on your side. Maybe `types-psutil` isn't installed correctly? It appears in `dev-requirements.txt`.
Try to remove this and run the CI again. If it passes I don't think we should add these lines.

YishaiZinkin (approved):
Cool

PR #17, opened 2021-12-13
Linux: ns: Return Process in get_mnt_ns_ancestor

YishaiZinkin:
A little API change - make get_mnt_ns_ancestor return a `Process` object instead of a pid.
Makes the function a little more generic and also will be used in https://github.com/Granulate/gprofiler/pull/240 to cache per namespace instead of per PID.

### DIFF
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -161,7 +161,7 @@ def get_proc_root_path(process: Process) -> str:
     """
     Gets /proc/<pid>/root of a given process, then a file can be read from the host mnt ns.
     """
-    return f"/proc/{get_mnt_ns_ancestor(process.pid)}/root"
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/ns.py
```diff
@@ -161,7 +161,7 @@ def get_proc_root_path(process: Process) -> str:
     """
     Gets /proc/<pid>/root of a given process, then a file can be read from the host mnt ns.
     """
-    return f"/proc/{get_mnt_ns_ancestor(process)}/root"
+    return f"/proc/{get_mnt_ns_ancestor(process.pid)}/root"
```
d3dave, line 164:
```suggestion
    return f"/proc/{get_mnt_ns_ancestor(process).pid}/root"
```

d3dave (changes requested):

d3dave (approved):

PR #20, opened 2022-01-16
linux: ns: get_process_nspid() support old kernels

adi-benz:
Make `get_process_nspid()` work with old kernels

Tested on Ubuntu 14.04 `3.13.0-170-generic`

### DIFF
--- a/granulate_utils/exceptions.py
+++ b/granulate_utils/exceptions.py
@@ -1,4 +1,4 @@
 class UnsupportedNamespaceError(Exception):
     def __init__(self, nstype: str):
-        super().__init__(f"Namespace '{nstype}' is not supported by this kernel")
+        super().__init__(f"Namespace {nstype!r} is not supported by this kernel")
         self.nstype = nstype
--- a/granulate_utils/exceptions.py
+++ b/granulate_utils/exceptions.py
@@ -1,0 +1,4 @@
+class UnsupportedNamespaceError(Exception):
+    def __init__(self, nstype: str):
+        super().__init__(f"Namespace '{nstype}' is not supported by this kernel")
+        self.nstype = nstype
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -100,7 +100,7 @@ def _get_process_nspid_by_sched_files(pid: int):
                 sched_file_path = procfs_child / "sched"
                 with sched_file_path.open("r") as sched_file:
                     sched_header_line = sched_file.readline()  # The first line contains the outer PID
-            except FileNotFoundError:
+            except (FileNotFoundError, ProcessLookupError):
                 # That's OK, processes might disappear before we get the chance to handle them
                 continue
 
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -13,6 +13,8 @@ from typing import Callable, List, Optional, TypeVar, Union
 
 from psutil import NoSuchProcess, Process
 
+from granulate_utils.exceptions import UnsupportedNamespaceError
+
 T = TypeVar("T")
 
 HOST_ROOT_PREFIX = "/proc/1/root"
@@ -156,22 +158,27 @@ def is_same_ns(process: Union[Process, int], nstype: str, process2: Union[Proces
     elif process2 is None:
         process2 = Process()  # `self`
 
-    is_the_same_ns = get_process_ns_inode(process, nstype) == get_process_ns_inode(process2, nstype)
-
-    # If one of the processes isn't running, we checked the wrong one
-    if not process.is_running():
-        raise NoSuchProcess(process.pid)
-    if not process2.is_running():
-        raise NoSuchProcess(process2.pid)
-
-    return is_the_same_ns
+    try:
+        return get_process_ns_inode(process, nstype) == get_process_ns_inode(process2, nstype)
+    except UnsupportedNamespaceError:
+        # The namespace does not exist in this kernel, hence the two processes are logically in the same namespace
+        return True
 
 
 def get_process_ns_inode(process: Process, nstype: str):
     try:
-        return os.stat(f"/proc/{process.pid}/ns/{nstype}").st_ino
+        ns_inode = os.stat(f"/proc/{process.pid}/ns/{nstype}").st_ino
     except FileNotFoundError as e:
-        raise NoSuchProcess(process.pid) from e
+        if process.is_running():
+            raise UnsupportedNamespaceError(nstype) from e
+        else:
+            raise NoSuchProcess(process.pid) from e
+
+    # If the process isn't running, we checked the wrong one
+    if not process.is_running():
+        raise NoSuchProcess(process.pid)
+
+    return ns_inode
 
 
 def run_in_ns(nstypes: List[str], callback: Callable[[], T], target_pid: int = 1) -> T:
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -68,35 +68,42 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
     return path
 
 
-def get_process_nspid(pid: int) -> int:
+def get_process_nspid(process: Union[Process, int]) -> int:
     """
     :raises NoSuchProcess: If the process doesn't or no longer exists
     """
-    nspid = _get_process_nspid_by_status_file(pid)
+    if isinstance(process, int):
+        process = Process(process)
+
+    nspid = _get_process_nspid_by_status_file(process)
     if nspid is not None:
         return nspid
 
-    if is_same_ns(pid, NsType.pid.name):
+    if is_same_ns(process, NsType.pid.name):
         # If we're in the same PID namespace, then the outer PID is also the inner pid (NSpid)
-        return pid
+        return process.pid
 
-    return _get_process_nspid_by_sched_files(pid)
+    return _get_process_nspid_by_sched_files(process)
 
 
-def _get_process_nspid_by_status_file(pid: int) -> Optional[int]:
+def _get_process_nspid_by_status_file(process: Process) -> Optional[int]:
     try:
-        with open(f"/proc/{pid}/status") as f:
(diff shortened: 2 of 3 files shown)


granulate_utils/linux/ns.py
```diff
@@ -67,6 +68,53 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+        for process_dir in procfs.iterdir():
+            is_process_dir = process_dir.is_dir() and process_dir.name.isdigit()
+            if not is_process_dir:
+                continue
+
+            try:
+                sched_file = process_dir / "sched"
+                sched_contents = sched_file.open("r").readline()
+                match = pattern.search(sched_contents)
+                if match:
```
Jongy, line 103:
```suggestion
                if match is not None:
```

granulate_utils/linux/ns.py
```diff
@@ -67,6 +68,53 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+    # Old kernel (pre 4.1) doesn't have an NSpid field in their /proc/pid/status file
+    # Instead, we can look through all /proc/*/sched files from inside the process' pid namespace, and due to a bug
+    # (fixed in 4.14) the outer PID is exposed, so we can find the target process by comparing the outer PID
+
+    def _find_inner_pid() -> Optional[int]:
+        pattern = re.compile(r"\((\d+), #threads: ")
+
+        procfs = Path("/proc")
+        for process_dir in procfs.iterdir():
+            is_process_dir = process_dir.is_dir() and process_dir.name.isdigit()
```
Jongy, line 95: Why is this expression stored in variable and not placed in the `if`?

granulate_utils/linux/ns.py
```diff
@@ -67,6 +68,53 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+    return _get_process_nspid_by_sched_files(pid)
+
+
+def _get_process_nspid_by_sched_files(pid: int):
+    # Old kernel (pre 4.1) doesn't have an NSpid field in their /proc/pid/status file
+    # Instead, we can look through all /proc/*/sched files from inside the process' pid namespace, and due to a bug
+    # (fixed in 4.14) the outer PID is exposed, so we can find the target process by comparing the outer PID
+
+    def _find_inner_pid() -> Optional[int]:
+        pattern = re.compile(r"\((\d+), #threads: ")
```
Jongy, line 91: Give an example to how one such line looks like? (always a good practice when writing regex - so the reader can see what you're trying to match)

granulate_utils/linux/ns.py
```diff
@@ -67,6 +68,53 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+
+        procfs = Path("/proc")
+        for process_dir in procfs.iterdir():
+            is_process_dir = process_dir.is_dir() and process_dir.name.isdigit()
+            if not is_process_dir:
+                continue
+
+            try:
+                sched_file = process_dir / "sched"
+                sched_contents = sched_file.open("r").readline()
```
Jongy, line 101: `with .. open(...)`?

Also - just `readline`, because this data is in the first line, right? Do comment

granulate_utils/linux/ns.py
```diff
@@ -67,6 +68,53 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+                match = pattern.search(sched_contents)
+                if match:
+                    outer_pid = int(match.group(1))
+                    if outer_pid == pid:
+                        return int(process_dir.name)
+            except FileNotFoundError:
```
Jongy, line 107: Preferably, wrap only the relevant code section with this try-except. So only the area that actually opens the file.

granulate_utils/linux/ns.py
```diff
@@ -67,6 +68,53 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+                # That's OK, processes might disappear before we get the chance to handle them
+                continue
+
+        return None
+
+    # We're searching `/proc`, so we only need to set our mount namespace
```
Jongy, line 113: Btw, some files in `/proc` do require changing netns/pidns, but specifically the numeric files for tasks - do not, only mntns

granulate_utils/linux/ns.py
adi-benz: For increased readability, this way the intention of the `if` is clearer IMO.

Jongy: Better done with a comment IMO, but okay as well

adi-benz:
Following your suggestions to use `assert`s (in https://github.com/Granulate/gprofiler/pull/269), now that `get_process_nspid()` fails only if the process isn't running, should we make the function raise instead of returning `None`?

Jongy:
> Following your suggestions to use `assert`s (in [Granulate/gprofiler#269](https://github.com/Granulate/gprofiler/pull/269)), now that `get_process_nspid()` fails only if the process isn't running, should we make the function raise instead of returning `None`?

Yes. Let's, instead of returning `None`, ensure that the PID now really doesn't run - and then raise `psutil.NoSuchProcess`. If it *does* run and we haven't found, raise another exception (that's probably a bug in our code)

granulate_utils/linux/ns.py
```diff
@@ -67,11 +68,84 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+                    return int(procfs_child.name)
+
+        return None
+
+    # We're searching `/proc`, so we only need to set our mount namespace
+    inner_pid = run_in_ns(["mnt"], _find_inner_pid, pid)
+    if inner_pid is not None:
+        return inner_pid
+
+    # If we weren't able to find the process' nspid, he must have been killed while searching
```
Jongy, line 134: Eh, this is racy.
1. Another PID may start meanwhile
2. PID might die and remain zombie, and this will return `True`.

What I suggest:
* Move https://github.com/Granulate/gprofiler/blob/8947923cd2a2560a248151e87d602b3460632cd5/gprofiler/utils.py#L483 to `granulate-utils`
* Make a `Process` object before calling `run_in_ns`, then check it afterwards?

adi-benz: I don't see a reason to use the `is_process_running()`.

`psutil`'s `is_running()` function is good in this case, I don't care if the process is zombie or not

Jongy: If the `sched` file exists for zombies - sure. If not, we will fail incorrectly, and then you do need to differentiate.

Jongy: Check and comment in code explaining

granulate_utils/linux/ns.py
```diff
@@ -67,11 +68,110 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+    if isinstance(process, int):
+        process = Process(process)
+    if isinstance(process2, int):
+        process2 = Process(process2)
+    elif process2 is None:
+        process2 = Process()  # `self`
+
+    is_the_same_ns = get_process_ns_inode(process, nstype) == get_process_ns_inode(process2, nstype)
+
+    # If one of the processes isn't running, we checked the wrong one
```
Jongy, line 161: Why do this check again, if already done in `get_process_ns_inode`?

granulate_utils/linux/ns.py
```diff
@@ -67,11 +68,110 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+    if not process.is_running():
+        raise NoSuchProcess(process.pid)
+    if not process2.is_running():
+        raise NoSuchProcess(process2.pid)
+
+    return is_the_same_ns
+
+
+def get_process_ns_inode(process: Process, nstype: str):
```
Jongy, line 170: Planning to export it? If not, prefix with `_`

granulate_utils/linux/ns.py
```diff
@@ -67,11 +68,110 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+    try:
+        return os.stat(f"/proc/{process.pid}/ns/{nstype}").st_ino
```
Jongy, line 172: One way this check can be improves: if we're checking for an `nstype` that doesn't exist in this kernel, this will raise, despite `is_same_ns`'s should actually treat it as "yes, they're the same namespace".

We should handle this case (e.g: `cgroup` namespace, not available in 3.10)

granulate_utils/linux/ns.py
adi-benz: `get_process_ns_inode()` doesn't make sure this is **the same** process.  
`get_process_ns_inode()` might return `True`, but actually, it was for the wrong process.

But now I'm moving this check into `get_process_ns_inode()`

granulate_utils/linux/ns.py
```diff
@@ -67,11 +70,115 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+
+    # We're searching `/proc`, so we only need to set our mount namespace
+    inner_pid = run_in_ns(["mnt"], _find_inner_pid, process.pid)
+    if inner_pid is not None:
+        if not process.is_running():  # Make sure the pid wasn't reused for another process
+            raise NoSuchProcess(process.pid)
+
+        return inner_pid
+
+    # If we weren't able to find the process' nspid, he must have been killed while searching (we only search
```
Jongy, line 146:
```suggestion
    # If we weren't able to find the process' nspid, it must have been killed while searching (we only search
```

granulate_utils/exceptions.py
```diff
@@ -0,0 +1,4 @@
+class UnsupportedNamespaceError(Exception):
+    def __init__(self, nstype: str):
+        super().__init__(f"Namespace '{nstype}' is not supported by this kernel")
```
Jongy, line 3:
```suggestion
        super().__init__(f"Namespace {nstype!r} is not supported by this kernel")
```

Jongy (changes requested):
.

Jongy (approved):

PR #23, opened 2022-01-19
Add process utils to handle zombie processes

ty93:
Added process utility functions to handle edge cases where processes are zombies


Jongy (approved):
LGTM. I'll update gProfiler later to use the new `is_process_running` function.

PR #24, opened 2022-01-27
Added `get_host_pid` function

IlayRosenberg:
Added a function that returns a process's host pid given its nspid and container id

### DIFF
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -11,7 +11,7 @@ from pathlib import Path
 from threading import Thread
 from typing import Callable, List, Optional, TypeVar, Union
 
-from psutil import NoSuchProcess, Process
+from psutil import NoSuchProcess, Process, process_iter
 
 from granulate_utils.exceptions import UnsupportedNamespaceError
 
@@ -282,3 +282,19 @@ def resolve_host_path(process: Process, ns_path: str) -> str:
     Get a path in the host mount namespace pointing to path in process mount namespace.
     """
     return resolve_proc_root_links(get_proc_root_path(process), ns_path)
+
+
+def get_host_pid(nspid: int, container_id: str) -> Optional[int]:
+    pid_namespace = ""
+    for process in process_iter():
+        try:
+            if not pid_namespace and container_id in Path(f"/proc/{process.pid}/cgroup").read_text():
+                pid_namespace = os.readlink(f"/proc/{process.pid}/ns/pid")
+            if pid_namespace and os.readlink(f"/proc/{process.pid}/ns/pid") == pid_namespace:
+                status = Path(f"/proc/{process.pid}/status").read_text()
+                for line in status.splitlines():
+                    if line.startswith("NSpid") and line.split()[-1] == str(nspid):
+                        return process.pid
+        except (FileNotFoundError, NoSuchProcess):
+            continue
+    return None
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -285,16 +285,15 @@ def resolve_host_path(process: Process, ns_path: str) -> str:
 
 
 def get_host_pid(nspid: int, container_id: str) -> Optional[int]:
+    assert len(container_id) == 64, f"Invalid container id {container_id!r}"
     pid_namespace = ""
     for process in process_iter():
         try:
             if not pid_namespace and container_id in Path(f"/proc/{process.pid}/cgroup").read_text():
                 pid_namespace = os.readlink(f"/proc/{process.pid}/ns/pid")
             if pid_namespace and os.readlink(f"/proc/{process.pid}/ns/pid") == pid_namespace:
-                status = Path(f"/proc/{process.pid}/status").read_text()
-                for line in status.splitlines():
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/ns.py
```diff
@@ -282,3 +282,19 @@ def resolve_host_path(process: Process, ns_path: str) -> str:
     Get a path in the host mount namespace pointing to path in process mount namespace.
     """
     return resolve_proc_root_links(get_proc_root_path(process), ns_path)
+
+
+def get_host_pid(nspid: int, container_id: str) -> Optional[int]:
```
Jongy, line 287: `assert` that `len(container_id) == 64`

granulate_utils/linux/ns.py
```diff
@@ -282,3 +282,19 @@ def resolve_host_path(process: Process, ns_path: str) -> str:
...
+    pid_namespace = ""
+    for process in process_iter():
+        try:
+            if not pid_namespace and container_id in Path(f"/proc/{process.pid}/cgroup").read_text():
+                pid_namespace = os.readlink(f"/proc/{process.pid}/ns/pid")
+            if pid_namespace and os.readlink(f"/proc/{process.pid}/ns/pid") == pid_namespace:
+                status = Path(f"/proc/{process.pid}/status").read_text()
```
Jongy, line 294: You can use `get_process_nspid` here instead of re-implementing it here, plus then it'll work for old kernels which don't have `NSpid`

Jongy (changes requested):
.

granulate_utils/linux/ns.py
```diff
@@ -282,3 +282,18 @@ def resolve_host_path(process: Process, ns_path: str) -> str:
...
     """
     return resolve_proc_root_links(get_proc_root_path(process), ns_path)
+
+
+def get_host_pid(nspid: int, container_id: str) -> Optional[int]:
+    assert len(container_id) == 64, f"Invalid container id {container_id!r}"
+    pid_namespace = ""
+    for process in process_iter():
+        try:
+            if not pid_namespace and container_id in Path(f"/proc/{process.pid}/cgroup").read_text():
```
Jongy, line 292: Here we should be using `_get_process_container_id` from gProfiler, I'll move to this repo and fix the code here; meanwhile we can merge this one.

granulate_utils/linux/ns.py
```diff
@@ -282,3 +282,18 @@ def resolve_host_path(process: Process, ns_path: str) -> str:
...
+                pid_namespace = os.readlink(f"/proc/{process.pid}/ns/pid")
+            if pid_namespace and os.readlink(f"/proc/{process.pid}/ns/pid") == pid_namespace:
```
Jongy, line 294: Could use `is_same_ns` but it'll make the code less concise... so up to you.

Jongy (approved):

Jongy:
@IlayRosenberg merging?

granulate_utils/linux/ns.py
```diff
@@ -282,3 +282,18 @@ def resolve_host_path(process: Process, ns_path: str) -> str:
...
     return resolve_proc_root_links(get_proc_root_path(process), ns_path)
+
+
+def get_host_pid(nspid: int, container_id: str) -> Optional[int]:
+    assert len(container_id) == 64, f"Invalid container id {container_id!r}"
+    pid_namespace = ""
+    for process in process_iter():
+        try:
+            if not pid_namespace and container_id in Path(f"/proc/{process.pid}/cgroup").read_text():
+                pid_namespace = os.readlink(f"/proc/{process.pid}/ns/pid")
```
Jongy, line 293: Technically, after starting, you need to re-start the loop: we want to catch processes that run in the PID NS but don't share the cgroup, each processes started by `nsenter` or `run_in_ns`.

granulate_utils/linux/ns.py
Jongy: Fixed in https://github.com/Granulate/granulate-utils/pull/25/files

granulate_utils/linux/ns.py
Jongy: Meh

Jongy (approved):

granulate_utils/linux/ns.py
```diff
@@ -282,3 +282,33 @@ def resolve_host_path(process: Process, ns_path: str) -> str:
...
+    assert len(container_id) == 64, f"Invalid container id {container_id!r}"
+
+    pid_namespace = ""
+    running_processes = list(process_iter())
+
+    # Get the pid namespace of the given container
+    for process in running_processes:
+        try:
+            if container_id in Path(f"/proc/{process.pid}/cgroup").read_text():
+                pid_namespace = os.readlink(f"/proc/{process.pid}/ns/pid")
```
Tamir-Mizrachi, line 297: You can add an optimization here - 
```code
            if container_id in Path(f"/proc/{process.pid}/cgroup").read_text():
                pid_namespace = os.readlink(f"/proc/{process.pid}/ns/pid")
                if get_process_nspid(process) == nspid:
                    return process.pid
                break
```
WDYT?

Jongy: https://github.com/Granulate/granulate-utils/pull/27/files#r795456428

PR #27, opened 2022-01-31
Minor optimization to get_host_pid

IlayRosenberg:


granulate_utils/linux/ns.py
```diff
@@ -295,6 +295,8 @@ def get_host_pid(nspid: int, container_id: str) -> Optional[int]:
         try:
             if container_id in Path(f"/proc/{process.pid}/cgroup").read_text():
                 pid_namespace = os.readlink(f"/proc/{process.pid}/ns/pid")
+                if os.readlink(f"/proc/{process.pid}/ns/pid") == pid_namespace and get_process_nspid(process) == nspid:
```
Tamir-Mizrachi, line 298: `os.readlink(f"/proc/{process.pid}/ns/pid") == pid_namespace` how it can be different value?

IlayRosenberg: Woops

granulate_utils/linux/ns.py
```diff
@@ -295,6 +295,8 @@ def get_host_pid(nspid: int, container_id: str) -> Optional[int]:
         try:
             if container_id in Path(f"/proc/{process.pid}/cgroup").read_text():
                 pid_namespace = os.readlink(f"/proc/{process.pid}/ns/pid")
+                if get_process_nspid(process) == nspid:
```
Jongy, line 298: I don't think it's worth the complication (not to mention that it's completely arbitrary whether this works or not - it depends on whether the target process is the first one we encounter from its cgroup...)

PR #36, opened 2022-03-06
linux: ns: Add full set of NsType

michelhe:


Jongy (approved):
Trust u on the numbers

PR #39, opened 2022-03-07
Feature/containers get container

ty93:
1. Added support for multiple cri runtimes existing at the same time, will now query all available cri sockets
2. Added support for `get_container(container_id: str)`

### DIFF
--- a/granulate_utils/containers/cri.py
+++ b/granulate_utils/containers/cri.py
@@ -4,13 +4,13 @@
 #
 
 import json
-from typing import List, Optional
+from typing import List, Optional, Union
 
 import grpc  # type: ignore # no types-grpc sadly
 
 import granulate_utils.generated.containers.cri.api_pb2 as api_pb2  # type: ignore
 from granulate_utils.containers.container import Container, ContainersClientInterface
-from granulate_utils.exceptions import CriNotAvailableError
+from granulate_utils.exceptions import ContainerNotFound, CriNotAvailableError
 from granulate_utils.generated.containers.cri.api_pb2_grpc import RuntimeServiceStub  # type: ignore
 from granulate_utils.linux.ns import resolve_host_root_links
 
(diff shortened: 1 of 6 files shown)


granulate_utils/containers/cri.py
```diff
@@ -70,29 +71,49 @@ def _reconstruct_name(self, container: api_pb2.Container) -> str:
...
-                        runtime=self._runtime_name,
-                        name=self._reconstruct_name(container),
-                        id=container.id,
-                        labels=container.labels,
-                        running=container.state == CONTAINER_RUNNING,
-                        pid=pid,
+                        api_pb2.ContainerStatusRequest(container_id=container_id, verbose=all_info)
                     )
-                )
+                except grpc._channel._InactiveRpcError:
```
Jongy, line 97: Why `except InactiveRpcError` here? (and if here, why not in `list_containers`?)

Jongy: And what if the container does not exist under containerd but does exist under cri-o?

Jongy (changes requested):
one

granulate_utils/containers/cri.py
ty93: You answered you own question :)
If the container doesn't exist under containerd it raises an `_InactiveRpcError` (A bit of a misleading name I think) with status = `StatusCode.NOT_FOUND`, and then we `continue` to try cri-o.

```
_InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.NOT_FOUND
	details = "an error occurred when try to find container "asdf": not found"
	debug_error_string = "{"created":"@1646903927.891701180","description":"Error received from peer unix:/run/containerd/containerd.sock","file":"src/core/lib/surface/call.cc","file_line":903,"grpc_message":"an error occurred when try to find container "asdf": not found","grpc_status":5}"
>

_InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.NOT_FOUND
	details = "could not find container "asdf": container with ID starting with asdf not found: ID does not exist"
	debug_error_string = "{"created":"@1646903802.171836819","description":"Error received from peer unix:/var/run/crio/crio.sock","file":"src/core/lib/surface/call.cc","file_line":903,"grpc_message":"could not find container "asdf": container with ID starting with asdf not found: ID does not exist","grpc_status":5}"
>
```

ty93: In `list_containers` we don't need it because we assume the connection is valid (since we checked it in the ctor) and there's no other reason for list_containers to fail

Jongy: I see. Indeed misleading, I knew that `InactiveRpcError` happens when the socket is closed. Can we filter the type and not continue if it's something else? i.e, `continue` only if `status == StatusCode.NOT_FOUND`?

ty93: Fixed

Jongy (approved):

PR #42, opened 2022-03-27
linux/ns: Support recursive symlinks in resolve_proc_root_links

ty93:
`resolve_proc_root_links` would check if some path is a link and resolve it, but then it wouldn't recheck if the resolved path is still a link. We now perform the check in a loop while keeping a set of checked paths to detect a symlink loop.
Note that while `os.path.realpath` and `os.path._joinrealpath` perform similar functionality, they won't prepend the `proc_root` prefix correctly after each link resolving.

### DIFF
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -72,10 +72,10 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
                 next_path = proc_root + link
             else:
                 # relative: just join
-                next_path = os.path.join(path, link)
+                next_path = os.path.join(os.path.dirname(next_path), link)
         path = next_path
 
-    return path
+    return os.path.abspath(path)
 
 
 def get_process_nspid(process: Union[Process, int]) -> int:
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -59,9 +59,13 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
     parts = Path(ns_path).parts
 
     path = proc_root
+    seen = set()
     for part in parts[1:]:  # skip the / (or multiple /// as .parts gives them)
         next_path = os.path.join(path, part)
-        if os.path.islink(next_path):
+        while os.path.islink(next_path):
+            if next_path in seen:
+                raise OSError("Symlink loop detected")
+            seen.add(next_path)
             link = os.readlink(next_path)
             if os.path.isabs(link):
                 # absolute - prefix with proc_root
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -64,7 +64,7 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
         next_path = os.path.join(path, part)
         while os.path.islink(next_path):
             if next_path in seen:
-                raise OSError("Symlink loop detected")
+                raise RuntimeError("Symlink loop from %r" % os.path.join(path, part))
             seen.add(next_path)
             link = os.readlink(next_path)
             if os.path.isabs(link):
@@ -75,7 +75,7 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
                 next_path = os.path.join(os.path.dirname(next_path), link)
         path = next_path
 
-    return os.path.abspath(path)
+    return path
 
 
 def get_process_nspid(process: Union[Process, int]) -> int:


granulate_utils/linux/ns.py
```diff
@@ -59,9 +59,13 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
     parts = Path(ns_path).parts
 
     path = proc_root
+    seen = set()
     for part in parts[1:]:  # skip the / (or multiple /// as .parts gives them)
         next_path = os.path.join(path, part)
-        if os.path.islink(next_path):
+        while os.path.islink(next_path):
```
d3dave, line 65: `path = next_path` also needs to be in the loop, I think.

d3dave (changes requested):

granulate_utils/linux/ns.py
Jongy: Seems so

ty93: I think you're right that what I did won't work for relative links, but I'm not sure `path = next_path` in the loop will fix it. I think maybe instead of using `path` for the relative case we should use `os.path.dirname(next_path)`, wdyt?

granulate_utils/linux/ns.py
```diff
@@ -59,19 +59,23 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+            if next_path in seen:
+                raise OSError("Symlink loop detected")
```
Jongy, line 67: I don't like this use of `OSError` (a standard exception with expected fields such as `errno`) for non standard errors. I think that `Exception` suffices here if we don't plan to have a specialized exception to help catching it.

granulate_utils/linux/ns.py
```diff
@@ -59,19 +59,23 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
                 # absolute - prefix with proc_root
                 next_path = proc_root + link
             else:
                 # relative: just join
-                next_path = os.path.join(path, link)
+                next_path = os.path.join(os.path.dirname(next_path), link)
         path = next_path
 
-    return path
+    return os.path.abspath(path)
```
Jongy, line 78: Why do you need `abspath` here? This function doesn't respect the concept of `/proc/pid/root`.

granulate_utils/linux/ns.py
Jongy: There's no need to get rid of `..`. The purpose of this function is to prepare a `/proc/pid/root` path that "the kernel will eat". Our only problem is with absolute symlinks - relative ones, incl `..` I suppose, work just fine.

Did you test it with a final path containing `..` and the resulting path turned out inaccessible?

Jongy (changes requested):
.

granulate_utils/linux/ns.py
ty93: This was to resolve `..` in the path, but if we don't care about that I can remove it

granulate_utils/linux/ns.py
ty93: I didn't have a specific problem with having `..` in the path, just thought that we can remove it like the corresponding `os.path.realpath` that does the same thing sans mounts. I'll remove it

granulate_utils/linux/ns.py
ty93: I can raise `OSError(errno.ELOOP)` which is raised from `os.path.realpath` if you want. I can also raise `RuntimeError("Symlink loop from %r" % e.filename)` which is what's raised from `pathlib.Path.resolve` as a wrapper for the `OSError`. The second one is probably better.

Jongy: We can go with the second one

granulate_utils/linux/ns.py
Jongy: Yup, can remove

granulate_utils/linux/ns.py
```diff
@@ -59,16 +59,20 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+                raise RuntimeError("Symlink loop from %r" % os.path.join(path, part))
+            seen.add(next_path)
             link = os.readlink(next_path)
             if os.path.isabs(link):
                 # absolute - prefix with proc_root
                 next_path = proc_root + link
             else:
                 # relative: just join
-                next_path = os.path.join(path, link)
+                next_path = os.path.join(os.path.dirname(next_path), link)
```
Jongy, line 75: I don't undersrand this part, what's the dirname for?

ty93: `next_path` is the file (or dir) that we're trying to resolve. So if it's a relative link, it's relative to the dir it resides in.
Example: if `/tmp/b` is a link to `a`, then it's a link to `/tmp/a`. If `/tmp/x/b` is a link to `../a` then it's a link to `/tmp/x/../a`.

Jongy: This basically means that relative links never worked, then. Right?

granulate_utils/linux/ns.py
Jongy: Current code lgtm, resolving this

granulate_utils/linux/ns.py
ty93: They did because we used to join `path` and not `next_path` (and `next_path` was originally just `join(path, part)` so `dirname(next_path) == path`. The reason I had to change it is because we can now resolve multiple times inside the loop and `dirname(next_path)` won't necessarily stay the same as `path` :)

Jongy: I see. Kk

Jongy (approved):

Jongy (approved):

ty93:
@YoniKF @Jongy Just a reminder that we need to decide if we want to merge this right now or if we want to change anything

Jongy:
> @YoniKF @Jongy Just a reminder that we need to decide if we want to merge this right now or if we want to change anything

I'm in favor of adding basic tests before continuing

ty93:
> > @YoniKF @Jongy Just a reminder that we need to decide if we want to merge this right now or if we want to change anything
> 
> I'm in favor of adding basic tests before continuing

I'm now in favor of merging and adding tests separately, @YoniKF @Jongy

Jongy:
We don't have any tests covering it in gProfiler as well. Merge now and I'll add one soon.

PR #44, opened 2022-03-31
Move run_process from gProfiler@1335d3f08cf

Jongy:
Move `run_process` and others from gProfiler to this repo.

See https://github.com/Granulate/gprofiler/pull/331

PR #45, opened 2022-04-11
Move utils/elf.py from gProfiler@9c7c02e3c39

ty93:


Jongy:
File is exactly the same as was in gprofiler?

Also, can you put in the commit msg the exact source of the file? So that it can be traced. Like I did in the title here: https://github.com/Granulate/granulate-utils/pull/44

ty93:
> File is exactly the same as was in gprofiler?
> 
> Also, can you put in the commit msg the exact source of the file? So that it can be traced. Like I did in the title here: #44

Exactly the same

Jongy:
Aw commit msg remained old

PR #47, opened 2022-04-24
Upgrade mypy to 0.942

shahar-gamliel:
Upgrade mypy to 0.942, as 0.910 is almost a year old.


Jongy (approved):
Great. Also update Granulate/gprofiler with the same version? (+ submodule bump)

PR #49, opened 2022-06-08
fix bug in case of broken exe path

shalomnaim1:
Adding a fix for cases that process exe is a broken link so instead of return an empty string as path, it raise an exception

### DIFF
--- a/granulate_utils/linux/process.py
+++ b/granulate_utils/linux/process.py
@@ -4,6 +4,12 @@
 #
 
 import psutil
+import os
+
+
+class DeletedExeException(psutil.Error):
+    def __init__(self, pid):
+        super(DeletedExeException, self).__init__(f"exe deleted for pid: {pid}")
 
 
 def process_exe(process: psutil.Process) -> str:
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/process.py
```diff
@@ -14,8 +20,13 @@ def process_exe(process: psutil.Process) -> str:
     See https://github.com/giampaolo/psutil/pull/2062
     """
     exe = process.exe()
-    if exe == "" and is_process_zombie(process):
-        raise psutil.ZombieProcess(process.pid)
+    if exe == "":
+        if is_process_zombie(process):
+            raise psutil.ZombieProcess(process.pid)
+        # following to https://man7.org/linux/man-pages/man5/proc.5.html,
+        # an empty exe can be returned in case of a broken link
```
Jongy, line 27: > If the pathname has been unlinked, the symbolic link will contain the string '(deleted)' appended to the original pathname.

That's what I'm seeing.

Also seeing this:

> In a multithreaded process, the contents of this symbolic link are not available if the main thread has already terminated

Which might be something that we encounter

granulate_utils/linux/process.py
```diff
@@ -4,6 +4,12 @@
 #
 
 import psutil
+import os
+
+
+class DeletedExeException(psutil.Error):
```
Jongy, line 10: `exceptions.py`

Jongy (changes requested):
.

PR #50, opened 2022-06-13
Add utility functions to read the contents of an elf symbol/address from disk

IlayRosenberg:


granulate_utils/linux/elf.py
```diff
@@ -96,3 +96,31 @@ def read_process_execfn(process: Process) -> str:
...
+def read_elf_symbol(path: str, sym_name: str, size: int) -> Optional[bytes]:
+    with open(path, "rb") as f:
+        elf = ELFFile(f)
+        symbols = elf.get_section_by_name(".symtab")
+        if symbols is None:
+            return None
+        symbol = symbols.get_symbol_by_name(sym_name)
+        if symbol is None:
+            return None
+        return _read_va_from_elf(elf, symbol[0].entry.st_value, size)
```
Jongy, line 126: `symbol[0]`? Is it an array of symbols?

IlayRosenberg: Yes, if there are different symbols with the same name

Jongy: Don't we want to ensure only one exists?

Jongy: Otherwise we might be returning the "incorrect" one

PR #52, opened 2022-06-15
added proxies to get request

marina-vrublevsky:
Using proxies (if found in env) in get request

### DIFF
--- a/granulate_utils/metadata/cloud.py
+++ b/granulate_utils/metadata/cloud.py
@@ -7,6 +7,7 @@ import logging
 from dataclasses import dataclass
 from http.client import NOT_FOUND
 from typing import Dict, List, Optional, Union
+from urllib.request import getproxies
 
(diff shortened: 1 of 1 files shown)


granulate_utils/metadata/cloud.py
```diff
@@ -136,7 +138,7 @@ def get_azure_metadata() -> Optional[AzureInstanceMetadata]:
 
 
 def send_request(url: str, headers: Dict[str, str] = None) -> Optional[Response]:
-    response = requests.get(url, headers=headers or {}, timeout=METADATA_REQUEST_TIMEOUT)
+    response = requests.get(url, headers=headers or {}, timeout=METADATA_REQUEST_TIMEOUT, proxies=getproxies())
```
Jongy, line 141: ? requests is doing it itself, doesn't it?

PR #53, opened 2022-06-16
Statically enforce ns types literal to avoid typos in nstypes in various callsites

michelhe:
I had a hard time debugging some code that uses `is_same_ns` after I found I used `mount` instead of `mnt` as the nstype :(

### DIFF
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -39,7 +39,7 @@ class NsType(enum.IntFlag):
 
 
 # note: keep in sync with the above NsType, duh.
-ValidNsTypeLiteral = Union[
+NsStr = Union[
     Literal["mnt"],
     Literal["net"],
     Literal["pid"],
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/ns.py
```diff
@@ -38,6 +38,18 @@ class NsType(enum.IntFlag):
     user = 0x10000000  # CLONE_NEWUSER
 
 
+# TODO: keep in sync with the above NsType, duh.
+ValidNsTypeLiteral = Union[
```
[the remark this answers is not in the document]
Jongy: Or maybe mypy doesn't do

Jongy (approved):

granulate_utils/linux/ns.py
```diff
@@ -38,6 +38,18 @@ class NsType(enum.IntFlag):
     user = 0x10000000  # CLONE_NEWUSER
 
 
+# TODO: keep in sync with the above NsType, duh.
```
Jongy, line 41:
```suggestion
# note: keep in sync with the above NsType, duh.
```

granulate_utils/linux/ns.py
d3dave, line 42:
```suggestion
NamespaceStr = Union[
```

d3dave (approved):
just a suggestion

granulate_utils/linux/ns.py
michelhe: `NsStr` like `NsType` above

d3dave (approved):

PR #56, opened 2022-07-04
cgroup support for memory and cpu limiting

marina-vrublevsky:
In this PR:

1. Add PID to memory and cpu cgroup
2. Limit memory of cgroup 
3. Limit amount of cores usage of cgroup

To Do:
1. Add memory and cpu usage methods
3. Support for cgroup v2


granulate_utils/linux/cgroups_client/memory_cgroup.py
```diff
@@ -0,0 +1,29 @@
...
+    HIERARCHY = "memory"
+
+    def get_memory_limit(self) -> int:
+        return int(self.read_from_controller("memory.limit_in_bytes"))
+
+    def get_max_memory(self) -> int:
+        return int(self.read_from_controller("memory.max_usage_in_bytes"))
+
+    def set_memory_limit(self, limit: int) -> None:
+
```
michelhe, line 20: remove newline

granulate_utils/linux/cgroups_client/base_cgroup.py
```diff
@@ -0,0 +1,93 @@
...
+
+PID_CGROUPS = Path("/proc/self/cgroup")
+CGROUP_PARENT_PATH = Path("/sys/fs/cgroup")
+
+IGNORE_LIST = ["kubepods", "docker", "ecs"]
+
+
+class HIERARCHIES(Enum):
+    memory = "memory"
+    cpu = "cpu,cpuacct"
```
michelhe, line 26: This enum seems not very much used (the values at least), do you have a particular use for it ?

michelhe (changes requested):
For now, not done

Jongy:
I'd like to review it as well, but first let's sync IRL on the purpose and what we try to achieve here, description wasn't clear enough to me

marina-vrublevsky:
> Generally I don't like the terminology of `cgroups_client`, I suggest simply `cgroups`, would be a good module name IMO

As i stated in the description (todo.2) there is currently another `cgroups.py` 
we need to decide if we want to merge them

granulate_utils/linux/cgroups_client/base_cgroup.py
marina-vrublevsky: https://github.com/Granulate/granulate-utils/pull/56/files#diff-85381afe3dc85f5e26b1f381a942c019c2b7830deaf508bc0e35805deb5a9a7bR63

michelhe:
> > Generally I don't like the terminology of `cgroups_client`, I suggest simply `cgroups`, would be a good module name IMO
> 
> @michelhe As i stated in the description (todo.2) there is currently another `cgroups.py` we need to decide if we want to merge them

It only consists of a single of function, you can just put it in the `__init__.py` of this module

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+
+import os
+from enum import Enum
+from pathlib import Path
+from typing import List
+
+from exceptions import MissingCgroup, MissingController, SkippedCgroup, UnsupportedCgroup
```
michelhe, line 11: this is not a relative import.
Anyway, we don't use relative imports, convert all to absolute

granulate_utils/linux/cgroups/__init__.py
```diff
@@ -10,6 +10,8 @@
 from .cpu_cgroup import CpuCgroup
 from .memory_cgroup import MemoryCgroup
 
+__all__ = ["CpuCgroup", "MemoryCgroup"]
```
michelhe, line 13: why?

marina-vrublevsky: removed

marina-vrublevsky: https://stackoverflow.com/questions/31079047/python-pep8-class-in-init-imported-but-not-used

granulate_utils/linux/cgroups/__init__.py
```diff
@@ -7,6 +7,9 @@
 
 from psutil import NoSuchProcess
 
+from .cpu_cgroup import CpuCgroup  # noqa: F401
+from .memory_cgroup import MemoryCgroup  # noqa: F401
+
```
michelhe, line 12: You use relative imports because absolute fails?

granulate_utils/linux/cgroups/__init__.py
```diff
@@ -7,6 +7,9 @@
 
 from psutil import NoSuchProcess
 
+from .cpu_cgroup import CpuCgroup  # noqa: F401
```
Jongy, line 10: Use absolute imports

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+
+class HIERARCHIES(Enum):
+    memory = "memory"
+    cpu = "cpu,cpuacct"
+
+
+class CgroupVerifications:
+    @staticmethod
+    def cgroup_v1() -> None:
+        if len(PID_CGROUPS.read_text().split("\n")) == 2:
```
Jongy, line 25: Explain here what do you expect for v1 and what for v2

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+            raise UnsupportedCgroup("version 2")
+
+    @staticmethod
+    def supported_cgroup(cgroup_type: str) -> None:
+        assert cgroup_type, "must provide cgroup type"
```
Jongy, line 30: Not necessary? An empty string is not in `HIERARCHIES.__members__`, don't need this extra check

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+        if cgroup_type not in HIERARCHIES.__members__:
+            raise UnsupportedCgroup(cgroup_type)
+
+    @staticmethod
+    def ignored_cgroup(cgroup: str, cgroup_name: str) -> None:
```
Jongy, line 35: "cgroup" and "cgroup_name" are not very good names. Both are practically the same

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+            parsed_cgroup = line.strip().split(":")
+            if HIERARCHIES[self.HIERARCHY].value == parsed_cgroup[1]:
+                return parsed_cgroup[2]
+
+        raise MissingCgroup(self.HIERARCHY, PID_CGROUPS.as_posix())
+
+    def move_to_cgroup(self, cgroup_name: str, pid: int = 0) -> None:
+        CgroupVerifications.ignored_cgroup(self.cgroup, cgroup_name)
+        os.makedirs(Path(self.cgroup_path / cgroup_name), exist_ok=True)
+        Path(self.cgroup_path / cgroup_name / "tasks").write_text("%s\n" % pid)
```
Jongy, line 65:
```suggestion
        Path(self.cgroup_path / cgroup_name / "tasks").write_text(str(pid))
```

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+        for line in cgroups.split("\n"):
+            parsed_cgroup = line.strip().split(":")
+            if HIERARCHIES[self.HIERARCHY].value == parsed_cgroup[1]:
+                return parsed_cgroup[2]
+
+        raise MissingCgroup(self.HIERARCHY, PID_CGROUPS.as_posix())
+
+    def move_to_cgroup(self, cgroup_name: str, pid: int = 0) -> None:
+        CgroupVerifications.ignored_cgroup(self.cgroup, cgroup_name)
+        os.makedirs(Path(self.cgroup_path / cgroup_name), exist_ok=True)
```
Jongy, line 64: Use `Path.makedirs` (or `mkdir`, whatever its name is)

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+PID_CGROUPS = Path("/proc/self/cgroup")
+CGROUP_PARENT_PATH = Path("/sys/fs/cgroup")
+
+
+class HIERARCHIES(Enum):
+    memory = "memory"
+    cpu = "cpu,cpuacct"
+
+
+class CgroupVerifications:
```
Jongy, line 22: The method names here are not imperative and its hard to understand what they mean. Either -
1. Change to imperative names, i.e `supported_cgroup` -> `is_cgroup_controller_supported`, `cgroup_v1` -> `is_cgroup_v1` etc.
2. Coalesce all functions into one `verify_preconditions` function

granulate_utils/linux/cgroups/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+        super().__init__(f"cgroup {cgroup_type} is unsupported in current release")
+
+
+class MissingCgroup(Exception):
+    def __init__(self, cgroup_type: str, pid_cgroup_file: str) -> None:
+        super().__init__(f"{cgroup_type} cgroup is missing from {pid_cgroup_file}")
+
+
+class MissingController(Exception):
+    def __init__(self, controller_path: Path) -> None:
```
Jongy, line 25: In other exceptions you pass `Path.as_posix()` but here you accept it and do the `as_posix`, be consistent

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class BaseCgroup:
+    HIERARCHY = ""
+
+    def __init__(self) -> None:
+        CgroupVerifications.cgroup_v1()
+        CgroupVerifications.supported_cgroup(self.HIERARCHY)
+
+    def _get_cgroup(self) -> str:
```
Jongy, line 53: This is practically the existing `get_cgroups` function, why not use it

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def read_from_controller(self, file_name: str) -> str:
+        controller = Path(self.cgroup_path / file_name)
+        CgroupVerifications.controller_exists(controller)
+        return controller.read_text()
+
+    def write_to_controller(self, file_name: str, data: str) -> None:
+        controller = Path(self.cgroup_path / file_name)
+        CgroupVerifications.controller_exists(controller)
+        controller.write_text(data)
```
Jongy, line 83:
```suggestion
        try:
            controller.write_text(data)
        except FileNotFoundError:
            raise ControllerNotFoundBlabla()
```

Instead of running one check and then doing one operation - prefer to run the operation and catch the appropriate error.

granulate_utils/linux/cgroups/cpu_cgroup.py
```diff
@@ -0,0 +1,21 @@
...
+#
+
+from granulate_utils.linux.cgroups.base_cgroup import BaseCgroup
+
+
+class CpuCgroup(BaseCgroup):
+    HIERARCHY = "cpu"
+
+    def set_cpu_limit_cores(self, cores: float) -> None:
+        period = int(self.read_from_controller("cpu.cfs_period_us").split("\n")[0])
```
Jongy, line 13:
```suggestion
        period = int(self.read_from_controller("cpu.cfs_period_us"))
```
it'll just work, `int` can handle whitespaces around the number. Same fix in other sites.

granulate_utils/linux/cgroups/cpu_cgroup.py
```diff
@@ -0,0 +1,21 @@
...
+        self.write_to_controller("cpu.cfs_quota_us", str(period * cores))
+
+    def get_cpu_limit_cores(self) -> float:
+        period = float(self.read_from_controller("cpu.cfs_period_us").split("\n")[0])
```
Jongy, line 17: Why `float`? It's an integer

granulate_utils/linux/cgroups/memory_cgroup.py
```diff
@@ -0,0 +1,28 @@
...
+    HIERARCHY = "memory"
+
+    def get_memory_limit(self) -> int:
+        return int(self.read_from_controller("memory.limit_in_bytes"))
+
+    def get_max_memory(self) -> int:
+        return int(self.read_from_controller("memory.max_usage_in_bytes"))
+
+    def set_memory_limit(self, limit: int) -> None:
+        self.write_to_controller("memory.limit_in_bytes", f"{limit}")
```
Jongy, line 20:
```suggestion
        self.write_to_controller("memory.limit_in_bytes", str(limit))
```
also below

granulate_utils/linux/cgroups/memory_cgroup.py
```diff
@@ -0,0 +1,28 @@
...
+from granulate_utils.linux.cgroups.exceptions import MissingController
+
+
+class MemoryCgroup(BaseCgroup):
+    HIERARCHY = "memory"
+
+    def get_memory_limit(self) -> int:
+        return int(self.read_from_controller("memory.limit_in_bytes"))
+
+    def get_max_memory(self) -> int:
```
Jongy, line 16: Let's use the exact names as cgroup uses to avoid confusion. So
```suggestion
    def get_max_usage_in_bytes(self) -> int:
```

granulate_utils/linux/cgroups/memory_cgroup.py
```diff
@@ -0,0 +1,28 @@
...
+        return int(self.read_from_controller("memory.max_usage_in_bytes"))
+
+    def set_memory_limit(self, limit: int) -> None:
```
Jongy, line 19:
```suggestion
    def set_limit_in_bytes(self, limit: int) -> None:
```

granulate_utils/linux/cgroups/memory_cgroup.py
```diff
@@ -0,0 +1,28 @@
...
+        self.write_to_controller("memory.limit_in_bytes", f"{limit}")
+        try:
+            self.write_to_controller("memory.memsw.limit_in_bytes", f"{limit}")
```
Jongy, line 22:
```suggestion
            self.write_to_controller("memory.memsw.limit_in_bytes", str(limit))
```

granulate_utils/linux/cgroups/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+
+from pathlib import Path
+
+
+class SkippedCgroup(Exception):
+    def __init__(self, current_cgroup: str, new_cgroup_name: str) -> None:
+        super().__init__(f"Skipping creating new {new_cgroup_name} cgroup since current cgroup is: {current_cgroup}")
```
Jongy, line 11:
```suggestion
        super().__init__(f"Skipping the creation of new {new_cgroup_name!r} cgroup since the current cgroup is {current_cgroup!r}")
```

granulate_utils/linux/cgroups/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class UnsupportedCgroup(Exception):
+    def __init__(self, cgroup_type: str) -> None:
+        super().__init__(f"cgroup {cgroup_type} is unsupported in current release")
```
Jongy, line 16:
```suggestion
        super().__init__(f"cgroup {cgroup_type!r} is unsupported")
```
"current release"?

granulate_utils/linux/cgroups/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class MissingCgroup(Exception):
+    def __init__(self, cgroup_type: str, pid_cgroup_file: str) -> None:
+        super().__init__(f"{cgroup_type} cgroup is missing from {pid_cgroup_file}")
```
Jongy, line 21:
```suggestion
        super().__init__(f"{cgroup_type!r} cgroup is missing from {pid_cgroup_file!r}")
```

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,87 @@
...
+from pathlib import Path
+from typing import List
+
+from granulate_utils.linux.cgroups.exceptions import MissingCgroup, MissingController, SkippedCgroup, UnsupportedCgroup
+
+PID_CGROUPS = Path("/proc/self/cgroup")
+CGROUP_PARENT_PATH = Path("/sys/fs/cgroup")
+
+
+class HIERARCHIES(Enum):
```
Jongy, line 17: Is the correct term here "hierarchies" or "subsystem"?

granulate_utils/linux/cgroups/cpu_cgroup.py
marina-vrublevsky: what if we want 1.5 cores?

Jongy: The `period` is always an integer. `/` operator is true divide even if both operands are integers

granulate_utils/linux/cgroups/base_cgroup.py
marina-vrublevsky: its a wrong name - ill replace with `CGROUPS`

granulate_utils/linux/cgroups/cpu_cgroup.py
marina-vrublevsky: right, i though you referred to something else

granulate_utils/linux/cgroups/base_cgroup.py
marina-vrublevsky: renamed `cgroup_name` -> `custom_cgroup`

granulate_utils/linux/cgroups/exceptions.py
marina-vrublevsky: meaning - currently version - "currently it doesn't support modifying X cgroup/cgroup v2"

granulate_utils/linux/cgroups/__init__.py
marina-vrublevsky: added absoulte

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,96 @@
...
+        for line in cgroups.split("\n"):
+            parsed_cgroup = line.strip().split(":")
+            if CGROUPS[self.controller].value == parsed_cgroup[1]:
+                return parsed_cgroup[2]
+
+        raise MissingCgroup(self.controller, PID_CGROUPS)
+
+    def move_to_cgroup(self, custom_cgroup: str, pid: int = 0) -> None:
+        CgroupVerifications.is_ignored_cgroup(self.cgroup, custom_cgroup)
+        Path.mkdir(Path(self.cgroup_path / custom_cgroup), exist_ok=True)
```
Jongy, line 68:
```suggestion
        Path(self.cgroup_path / custom_cgroup).mkdir(exist_ok=True)
```

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,96 @@
...
+    def read_from_controller(self, file_name: str) -> str:
+        controller_path = Path(self.cgroup_path / file_name)
+        try:
+            return controller_path.read_text()
+        except FileNotFoundError:
+            raise MissingController(controller_path)
+
+    def write_to_controller(self, file_name: str, data: str) -> None:
+        controller_path = Path(self.cgroup_path / file_name)
+        CgroupVerifications.is_controller_exists(controller_path)
```
Jongy, line 88: Don't need this. Exception will be raised by `write_text` if it doesn't exist

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,96 @@
...
+        try:
+            controller_path.write_text(data)
+        except FileNotFoundError:
+            raise MissingController(controller_path)
+
+    def get_cgroup_pids(self) -> List[str]:
+        content = Path(self.cgroup_path / "tasks").read_text()
+        return list(filter(None, content.split("\n")))
```
Jongy, line 96: Cool trick, didn't know of `filter(None, ...)`.

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def write_to_controller(self, file_name: str, data: str) -> None:
+        controller_path = Path(self.cgroup_path / file_name)
+        CgroupVerifications.is_controller_exists(controller_path)
+        try:
+            controller_path.write_text(data)
+        except FileNotFoundError:
+            raise MissingController(controller_path)
+
+    def get_cgroup_pids(self) -> List[str]:
```
Jongy, line 94: Makes more sense to return `List[int]` for PIDs

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,96 @@
...
+#
+
+from enum import Enum
+from pathlib import Path
+from typing import List
+
+from granulate_utils.linux.cgroups.exceptions import MissingCgroup, MissingController, SkippedCgroup, UnsupportedCgroup
+
+PID_CGROUPS = Path("/proc/self/cgroup")
+CGROUP_PARENT_PATH = Path("/sys/fs/cgroup")
```
Jongy, line 13: CGROUPFS is the right name here (the cgroup fs is mounted on `/sys/fs/cgroup` and is possibly mounted on other directories, but it's okay to support only this path for MVP)

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,96 @@
...
+        if len(PID_CGROUPS.read_text().split("\n")) == 2:
+            raise UnsupportedCgroup("version 2")
+
+    @staticmethod
+    def is_supported_cgroup(cgroup_type: str) -> None:
+        if cgroup_type not in CGROUPS.__members__:
+            raise UnsupportedCgroup(cgroup_type)
+
+    @staticmethod
+    def is_ignored_cgroup(cgroup: str, custom_cgroup: str) -> None:
```
Jongy, line 37: This is actualy the opposite - unlike `is_supported_cgroup` which asserts the subsystem exists, this one asserts that it is NOT ignored

Jongy (changes requested):
I'll complete the pass later. Meanwhile - we still have naming weirdnesses here, so let's define clear names, following https://man7.org/linux/man-pages/man7/cgroups.7.html:

* A cgroup type i.e `memory`, `cpu,cpuacct` is called `subsystem`. Ensure all `cgroup_type` and `controller` are renamed to `subsystem`. (You may also choose `controller`, I don't mind, but be consistent)
* A cgroup path (i.e `/`, `/docker/....`) is called `cgroup`
* Any other?

granulate_utils/linux/cgroups/cpu_cgroup.py
```diff
@@ -0,0 +1,21 @@
...
+#
+
+from granulate_utils.linux.cgroups.base_cgroup import BaseCgroup
+
+
+class CpuCgroup(BaseCgroup):
+    controller = "cpu"
+
+    def set_cpu_limit_cores(self, cores: float) -> None:
+        period = int(self.read_from_controller("cpu.cfs_period_us"))
```
Jongy, line 13: File paths i.e `"cpu.cfs_quota_us"` can be consitified and reused in this class

granulate_utils/linux/cgroups/memory_cgroup.py
```diff
@@ -0,0 +1,28 @@
...
+        return int(self.read_from_controller("memory.limit_in_bytes"))
+
+    def get_max_usage_in_bytes(self) -> int:
+        return int(self.read_from_controller("memory.max_usage_in_bytes"))
+
+    def set_limit_in_bytes(self, limit: int) -> None:
+        self.write_to_controller("memory.limit_in_bytes", str(limit))
+        try:
+            self.write_to_controller("memory.memsw.limit_in_bytes", str(limit))
+        except MissingController:
```
Jongy, line 23: Exception is confusing here - it's not that the memory controller is missing, but the specific file is missing. We can split it to 2 exceptions - one is `MissingSubsystem` for when you try to use `cpu,cpuacct` e.g and that's non existent, and one when you try to use a file in a subsystem and the file itself isn't present - that's another error that needs to be distinguished.

granulate_utils/linux/cgroups/memory_cgroup.py
```diff
@@ -0,0 +1,28 @@
...
+    controller = "memory"
+
+    def get_memory_limit(self) -> int:
+        return int(self.read_from_controller("memory.limit_in_bytes"))
+
+    def get_max_usage_in_bytes(self) -> int:
+        return int(self.read_from_controller("memory.max_usage_in_bytes"))
+
+    def set_limit_in_bytes(self, limit: int) -> None:
+        self.write_to_controller("memory.limit_in_bytes", str(limit))
```
marina-vrublevsky, line 20: omg forgot to remove this

granulate_utils/linux/cgroups/memory_cgroup.py
```diff
@@ -24,7 +24,6 @@ def set_limit_in_bytes(self, limit: int) -> None:
             self.write_to_control_file(self.memsw_limit_in_bytes, str(limit))
         except PermissionError:
             # if swap extension is not enabled (CONFIG_MEMCG_SWAP) this file doesn't exist
-            # and PermissionError is thrown (since it can't be created)
```
Jongy: Why removed? this makes sense to add here

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # "/proc/$PID/cgroup" lists a process's cgroup membership.  If legacy
+        # cgroup is in use in the system, this file may contain multiple lines, one for each hierarchy.
+        # The entry for cgroup v2 is always in the format "0::$PATH"::
+        if len(PID_CGROUPS.read_text().split("\n")) == 2:
+            raise Exception("cgroup V2 is unsupported")
+
+    @property
+    def subsystem(self) -> str:
+        raise NotImplementedError()
```
Jongy, line 36:
```suggestion
        raise NotImplementedError
```
nitpick

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def _get_cgroup(self) -> str:
+        for line in PID_CGROUPS.read_text().split("\n"):
+            hierarchy = line.strip().split(":")
+            if SUBSYSTEMS[self.subsystem].value == hierarchy[1]:
+                return hierarchy[2]
+        raise Exception(f"{self.subsystem!r} is not in found")
```
Jongy, line 43:
```suggestion
        raise Exception(f"{self.subsystem!r} not found")
```

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def move_to_cgroup(self, custom_cgroup: str, pid: int = 0) -> None:
```
Jongy, line 45:
```suggestion
    def move_to_cgroup(self, custom_cgroup: str, tid: int = 0) -> None:
```

logically, threads are part of cgroups, not processes

Jongy: Document that if `tid==0`, Linux will use the current thread

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,69 @@
...
+        if any(x in self.predefined_cgroups for x in self.cgroup.split("/")):
+            raise Exception(f"{self.subsystem!r} subsytem is already in a predefined cgroup: {self.cgroup!r}")
+        Path(self.cgroup_mount_path / custom_cgroup).mkdir(exist_ok=True)
+        Path(self.cgroup_mount_path / custom_cgroup / "tasks").write_text(str(pid))
```
Jongy, line 49: `cgroup_mount_path` includes the current cgroup path, so if I'm in `/a` and `custom_cgroup = 'b'` you'll move me to `/a/b`, no?

marina-vrublevsky: yeah - the logic here is to move to a new cgroup inside the current cgroup

granulate_utils/linux/cgroups/base_cgroup.py
Jongy: I disagree - well, it's more important to have logic written in one site and not duplicated. You can write a function here `_get_cgroup` that calls the other one. You can change the other one to call this function. I don't mind, but don't duplicate logic

granulate_utils/linux/cgroups/base_cgroup.py
Jongy: Right. Recalled now. Document pls

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -0,0 +1,69 @@
...
+    def _get_cgroup(self) -> str:
+        for line in PID_CGROUPS.read_text().split("\n"):
+            hierarchy = line.strip().split(":")
+            if SUBSYSTEMS[self.subsystem].value == hierarchy[1]:
+                return hierarchy[2]
+        raise Exception(f"{self.subsystem!r} is not in found")
+
+    def move_to_cgroup(self, custom_cgroup: str, pid: int = 0) -> None:
+        if any(x in self.predefined_cgroups for x in self.cgroup.split("/")):
+            raise Exception(f"{self.subsystem!r} subsytem is already in a predefined cgroup: {self.cgroup!r}")
```
Jongy, line 47: Upon usage - this exception needs to be caught by this API's users, to decide to bail on cgroup changes. Therefore this one makes sense as a specific type exception, to allow easily catching it

Jongy (changes requested):
Few last. Please test it as we discussed in private

Jongy (approved):

PR #57, opened 2022-07-26
Move process-related functions from elf.py to process.py

d3dave:
Process-related functions were erroneously placed in elf.py. Moved them to process.py instead.


Jongy (approved):
Trust you it's just moving around <3

PR #58, opened 2022-07-31
added region to gcp cloud

tomersa:


granulate_utils/metadata/cloud.py
```diff
@@ -103,6 +107,13 @@ def get_gcp_metadata() -> Optional[GcpInstanceMetadata]:
     )
 
 
+def get_gcp_region_from_zone(zone: str) -> str:
+    zone = zone.rsplit('/', 1)[-1]
```
Jongy, line 111: What's the operation you are doing here? How does it look exactly? Give a sample comment on input/output when doing string manipulations

granulate_utils/metadata/cloud.py
```diff
@@ -44,6 +44,7 @@ class GcpInstanceMetadata(InstanceMetadataBase):
     instance_id: str
     image_id: str
     name: str
+    region: str = None
```
Jongy, line 47: Why `= None`?

granulate_utils/metadata/cloud.py
tomersa: An example of zone would be: `projects/234/zones/us-east1-a`
So,
`get_gcp_region_from_zone('projects/234/zones/us-east1-a')` -> returns
`us-east1`

granulate_utils/metadata/cloud.py
tomersa: Actually, I can remove it. It's a leftover.

granulate_utils/metadata/cloud.py
Jongy: Comment in code pls

Jongy (changes requested):
Just fix lint

ziv-bar (changes requested):
Nice fix, please take a look at the comments below :)
1. This fix commits the change in sAgent. This means we will have to upgrade every `gcp` sAgent now, which I understand is a big deal.
2. The original `zone` parsing happened in `gServer` (probably for the reason above), I suggest we move the logic to `dServer` instead of sAgent.
3. The `gServer` cloud info code does two parse operations to `zone`:
gserver_cloud_metadata_parsing png
* Remove the "project..." prefix from`zone`
* Create the `region` key from the new zone.
I think that if we make a change, we should implement both of the parse operations on `zone`... Same way as the gServer.

tomersa:
> Nice fix, but perhaps the change should be made in a different place.

As discussed given the דיחפות of the situation we will solve it in here.

Jongy:
I think this is the right place to add - this way it's also available to gProfiler

tomersa:
After a discussion with @Tamir-Mizrachi I'm going to move it to dServer. So ignore closing this PR

PR #60, opened 2022-08-08
cloud: Use IMDSv2 to query AWS metadata

Jongy:


Jongy:
Tested gProfiler built with it on an instance accepting IMDSv1 and IMDSv2 (where the previous version worked as well) and on an instance accepting only IMDSv2. Works in both cases.

I couldn't find any instance accepting only v1 so I think we can drop support for that (althoguh the documentation I refered to here claims that "AWS SDK will try v1 if v2 fails" but I didn't find any reference to that in modern botocore code - it tries only v2 :shrug:).

granulate_utils/linux/process.py
```diff
@@ -4,13 +4,13 @@
 #
 
 import struct
-import psutil
```
Jongy, line 7: lint stuff, IDK why I got them and CI didn't catch

michelhe (approved):

PR #62, opened 2022-08-14
Introducing gLogger SDK

d3dave:
Added a `logging.Handler` named `BatchRequestHandler` that accumulates dicts that represent log messages. Messages are flushed asynchronously in a separate thread every interval or when a threshold is reached. Flushing sends the logs using a POST request to a web server. Messages are retained until an affirmative response from the server is received. If the total length of all messages exceeds the defined maximum, then the oldest messages are discarded to provide space for the new messages to be written. This happens synchronously during emission (not on the flush thread), to prevent accidental ballooning of the buffer should the flush thread crash/freeze. 

To prevent old messages from being discarded prematurely before they are sent, we must ensure that we try to flush before ever reaching the maximum capacity (because discarding happens during emission, and not on the flush thread). That's why the flush threshold should be set at some reasonable percentage below 100. In addition, there's an exponential backoff to send failures that retries before giving up. The retries count is configurable and is 5 times by default.

To support the max length limit functionality, all messages are stored serialized in a `MessagesBuffer`. This class tracks the total length of all items currently stored in it.


d3dave:
I'll take a look at the tests later.

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+class BatchingHandler(Handler):
+    """
+    logging.Handler that accumulates logs in a buffer and flushes them upon explicit request.
+    The logs are transformed into dicts.
+    """
+    capacity = 100 * 1000  # max number of records to buffer
+
+    def __init__(self) -> None:
+        super().__init__(logging.DEBUG)
+        self._buffer: list[dict] = []
```
Jongy, line 20: For gProfiler don't assume Python 3.9 types, it runs with older as well

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        Called when amount of records in buffer exceeds the capacity.
+        Default implementation does nothing.
+        """
+        pass
+
+    def flush(self) -> None:
+        # Snapshot the current num logs because logs list might be extended meanwhile.
+        logs_count = len(self._buffer)
+        self.flush_logs(self._buffer[:logs_count])
+        # If succeeded, remove the sent logs from the list.
```
Jongy, line 43:
```suggestion
        # If succeeded, remove the flushed logs from the list.
```

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        # The formatter is needed to format tracebacks
+        self.setFormatter(logging.Formatter())
+
+    def emit(self, record: LogRecord) -> None:
+        item = self.dictify_record(record)
+        self._buffer.append(item)
+        # truncate logs to last N entries
+        if len(self._buffer) > self.capacity:
+            self.on_truncated()
+            self._buffer[: -self.capacity] = []
```
Jongy, line 30: This removes the first N extra items, right? This means `on_truncated` will be called very often if there are errors. I think an API like the way the gProfiler class works makes more sense (i.e pass `was_truncated` when calling `flush_logs`. Or call `on_truncated()` once per `flush_logs` call?

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+            "level": record.levelname,
+            "timestamp": formatted_timestamp,
+            "logger_name": record.name,
+            "exception": exception_traceback,
+        }
+
+    def _format_exception(self, record: LogRecord) -> str:
+        assert self.formatter is not None
+        if record.exc_info:
+            # Use cached exc_text if available.
```
Jongy, line 63: "cached exc_text"? What's caching it?

Jongy (dismissed):
.

granulate_utils/logging/__init__.py
d3dave: Hmm... yes, as a matter of fact most of the time (ie. non-concurrent calls) this would probably remove just one item from the head of the list consistently until flushed. A better option would probably be to toss $k \times Capacity$ on overflow.

d3dave:
@Jongy Updated with threaded implementation.

d3dave:
Here's an example request:

```
POST / HTTP/1.1
Host: localhost
User-Agent: python-requests/2.27.1
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
Content-Length: 1325
Content-Type: application/json

{"id": "9c8e8244788846b7a6f04682cb08f23c", "metadata": {}, "logs": [{"serial_no": 2, "message": "test 2", "level": "ERROR", "timestamp": "2022-08-16T20:20:35.957511", "logger_name": "root"}, {"serial_no": 3, "message": "test 3", "level": "ERROR", "timestamp": "2022-08-16T20:20:35.957526", "logger_name": "root"}, {"serial_no": 4, "message": "test 4", "level": "ERROR", "timestamp": "2022-08-16T20:20:35.957540", "logger_name": "root"}, {"serial_no": 5, "message": "test 5", "level": "ERROR", "timestamp": "2022-08-16T20:20:35.957553", "logger_name": "root"}, {"serial_no": 6, "message": "test 6", "level": "ERROR", "timestamp": "2022-08-16T20:20:35.957566", "logger_name": "root"}, {"serial_no": 7, "message": "test 7", "level": "ERROR", "timestamp": "2022-08-16T20:20:35.957578", "logger_name": "root"}, {"serial_no": 8, "message": "test 8", "level": "ERROR", "timestamp": "2022-08-16T20:20:35.957591", "logger_name": "root"}, {"serial_no": 9, "message": "test 9", "level": "ERROR", "timestamp": "2022-08-16T20:20:35.957606", "logger_name": "root"}, {"serial_no": 10, "message": "Backing off _send_logs(...) for 0.1s (requests.exceptions.ReadTimeout: HTTPConnectionPool(host='localhost', port=80): Read timed out. (read timeout=1.5))", "level": "INFO", "timestamp": "2022-08-16T20:20:38.481337", "logger_name": "backoff"}]}
```

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+        except Exception:
+            logger.error("Error posting to server")
+        else:
+            self.remove_up_to_serial_no(last_ingested_serial_no + 1)
+
+    @backoff.on_exception(backoff.expo, exception=RequestException, max_tries=REQUEST_MAX_TRIES)
+    def _send_logs(self) -> requests.Response:
+        # Upon every retry we will remake the batch, in case we are able to batch more messages together.
+        data = {
+            "id": uuid.uuid4().hex,
```
rgruner, line 107: The `id` is named differently in the server - please sync with Sasson

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+            "metadata": self.get_metadata(),
+            "logs": self.make_batch(),
+        }
+        return self.session.post(f"{self.scheme}://{self.server_address}/", json=data, timeout=self.request_timeout)
+
+    def get_metadata(self) -> dict:
+        """Called to get metadata per batch."""
+        return {}
```
rgruner, line 115: Need to implement basic ones and be able to augment with additional metadata from the ctor (so that agents will be able to add more custom metadata)

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class BatchRequestsHandler(Handler):
+    """
+    logging.Handler that accumulates logs in a buffer and flushes them periodically or when
+    threshold is reached. The logs are transformed into dicts. Requests to the server are serialized
+    in a dedicated thread to reduce congestion on the server.
+    """
+
+    scheme = "https"
```
rgruner, line 29: Can't these be defaults that can be specified in the handler ctor?

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+import backoff
+import requests
+from requests import RequestException
+
+REQUEST_MAX_TRIES = 5
+
+logger = logging.getLogger(__name__)
+
+
+class BatchRequestsHandler(Handler):
```
rgruner, line 22: How is this used? Would add documentation on how to use it and maybe even create a function to help this process **if** its complicated

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+import uuid
+from datetime import datetime
+from logging import Handler, LogRecord
+from typing import Any, Dict, List, Tuple, Union
+
+import backoff
+import requests
+from requests import RequestException
+
+REQUEST_MAX_TRIES = 5
```
rgruner, line 17: Allow control through env variable?

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,138 @@
+#
```
rgruner, line 1: Would rename folder to `glogger-sdk` and put it in its own file instead of __init__.py (unless you have a good reason to put it here)

rgruner (changes requested):

rgruner (changes requested):

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,230 @@
...
+        flush_interval=10.0,
+        flush_threshold=0.8,
+        overflow_drop_factor=0.25,
+    ):
+        super().__init__(logging.DEBUG)
+        self.max_message_size = max_message_size  # maximum message size
+        self.capacity = max_total_length  # maximum size of buffer in bytes
+        self.flush_interval = flush_interval  # maximum amount of seconds between flushes
+        self.flush_threshold = flush_threshold  # force flush if buffer size reaches this percentage of capacity
+        self.overflow_drop_factor = overflow_drop_factor  # drop this percentage of messages upon overflow
```
d3dave, line 106: Remove from here, it's only used in MessagesBuffer

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,230 @@
...
+        :return: (batch that was sent, response)
+        """
+        # Upon every retry we will remake the batch, in case we are able to batch more messages together.
+        batch = self.make_batch()
+        data = {
+            "id": uuid.uuid4().hex,
+            "metadata": self.get_metadata(),
+            "logs": "<LOGS_JSON>",
+        }
+        # batch.logs is a list of strings so ",".join() it into the final json string instead of json-ing the list.
```
d3dave, line 201:
```suggestion
        # batch.logs is a list of json.dump()ed strings so ",".join() it into the final json string instead of json-ing the list.
```

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,230 @@
...
+        self.time_fn = time.time
+        self.server_address = server_address
+        self.session = requests.Session()
+        self.last_flush_time = 0.0
+        self.flush_thread = threading.Thread(target=self._flush_loop, daemon=True, name="Logs flusher")
+        self.start()
+
+    def start(self) -> None:
+        """Called to start the flush thread."""
+        self.flush_thread.start()
```
d3dave, line 118: Should we recreate the thread in `start`? IMO It's not expected to be restarted. If restart is desired, recreate the entire handler. Otherwise, we'd be invalidating some invariants and dropping messages for no reason. Comment this.

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,230 @@
...
+            "logger_name": record.name,
+            "message": record.message,
+        }
+
+    def _flush_loop(self) -> None:
+        self.last_flush_time = self.time_fn()
+        while not self.stop_event.is_set():
+            if self.should_flush():
+                self.flush()
+            self.stop_event.wait(1.0)
```
d3dave, line 158: This can be provided in a mixin, but it's not important for now.

granulate_utils/logging/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+        if len(self.buffer) >= self.capacity:
+            drop_count = int(self.overflow_drop_factor * self.capacity)
+            del self.buffer[:drop_count]
+
+    def dictify_record(self, record: LogRecord) -> Dict[str, Any]:
+        # `format` is required to fill in some record members such as `message` and `asctime`.
+        self.format(record)
+        formatted_timestamp = datetime.utcfromtimestamp(record.created).isoformat()
+        return {
+            "serial_no": self.next_serial_no,
```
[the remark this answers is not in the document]
d3dave: If `emit` is called directly, yes. But it's supposed to be called only from `Handler.handle` which holds the lock while calling `emit`.

granulate_utils/logging/__init__.py
d3dave: Most basic usage would be something like this:
```python
handler = BatchRequestsHandler("localhost:61234")
logging.basicConfig(handlers=[handler])
```

More fine-grained would probably look like this:
```python
logger = logging.getLogger("gprofiler")
handler = BatchRequestsHandler("localhost:61234")
handler.setLevel(logging.DEBUG)
logger.addHandler(handler)
```

granulate_utils/logging/__init__.py
rgruner: No need, but worth to add assert to make sure it is not recreated (and add as test)

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+        self,
+        server_address: str,
+        max_message_size=1 * 1024 * 1024,  # 1mb
+        max_total_length=5 * 1024 * 1024,  # 5mb
+        flush_interval=10.0,
+        flush_threshold=0.8,
+        overflow_drop_factor=0.25,
+        max_send_tries=5,
+    ):
+        super().__init__(logging.DEBUG)
```
rgruner, line 55: Why DEBUG? Allow to control?

tests/test_glogger.py
```diff
@@ -0,0 +1,233 @@
...
+        logging.basicConfig(force=True, handlers=[handler], level=0)
+        logging.info("A" * 1000)
+        logging.info("B" * 2000)
+        logging.info("C" * 3000)
+        logging.info("D" * 4000)
+        logs_server.handle_request()
+        assert logs_server.processed > 0
+        last_message = json.loads(handler.messages_buffer.buffer[-1])
+        assert last_message["severity"] == logging.ERROR
+        assert last_message["message"] == "Error posting to server"
```
rgruner, line 123: Use a constant for it.

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+        send = backoff.on_exception(backoff.expo, exception=RequestException, max_tries=self.max_send_tries)(
+            self._send_logs
+        )
+
+        self.last_flush_time = self.time_fn()
+        try:
+            batch, response = send()
+            response.raise_for_status()
+        except Exception:
+            logger.exception("Error posting to server")
```
rgruner, line 136: This really scares me that it can cause issues in multithreading - WDYT?

tests/test_glogger.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def test_truncate_long_message():
+    """Test message is truncated and marked accordingly if it's longer than max message size."""
+    with ExitStack() as exit_stack:
```
rgruner, line 128: The exitStack appears to many time - I suggest to use pytest fixtures to wrap this up more nicely

tests/test_glogger.py
```diff
@@ -0,0 +1,233 @@
...
+        # we don't need a real port for this one
+        handler = HttpBatchRequestsHandler("localhost:61234", max_message_size=1000)
+        exit_stack.callback(handler.stop)
+
+        logging.basicConfig(force=True, handlers=[handler], level=0)
+        logging.info("A" * 2000)
+        assert_buffer_attributes(handler, count=1)
+        s = handler.messages_buffer.buffer[0]
+        m = json.loads(s)
+        assert len(m["message"]) <= 1000
```
rgruner, line 138: This is the wrong test, everything that is sent to the server should be less than the amount (that's how the server parses it as it doesn't parse json that is longer than X)

tests/test_glogger.py
```diff
@@ -0,0 +1,233 @@
...
+
+        for t in threads:
+            t.start()
+
+        for t in threads:
+            t.join()
+
+        assert session.posts > 5
+        assert session.successes == session.posts
+        assert session.errors == 0
```
rgruner, line 233: Missing a test on retry logic (that attempts to try X times before dropping the buffer - need to simulate and test this flow somehow)

rgruner (changes requested):

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+        record_dict = self.dictify_record(record)
+        if record.truncated:  # type: ignore
+            record_dict["truncated"] = True
+        record_dict["serial_no"] = self.messages_buffer.next_serial_no
+        return json.dumps(record_dict)
+
+    def format(self, record: LogRecord) -> str:
+        """Override to customize record fields."""
+        s = super().format(record)
+        if len(record.message) > self.max_message_size:
```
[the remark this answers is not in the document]
d3dave: In such a case what should we do? Truncate the message to ~75% of max length?

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+        self.flush_thread.start()
+
+    def emit(self, record: LogRecord) -> None:
+        # Called while lock is acquired
+        item = self.dump_record_locked(record)
+        self.messages_buffer.append(item)
+
+    def dump_record_locked(self, record: LogRecord) -> str:
+        # `format` is required to fill in some record members such as `message` and `asctime`.
+        self.format(record)
```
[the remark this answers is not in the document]
d3dave: The intent is for a subclass to be able to override `dictify_record` to add fields to the dict, and override `format` to add fields to the record, both without needing to copy the `json.dumps` and the `truncated` and `serial_no` fields.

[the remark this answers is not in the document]
d3dave: If we place the logic in `format`, then a subclass needs to do `json.dumps` and re-add the fields if it wants to make any changes to the record dict.

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+    def _send_logs(self) -> Tuple[Batch, requests.Response]:
+        """
+        :return: (batch that was sent, response)
+        """
+        # Upon every retry we will remake the batch, in case we are able to batch more messages together.
+        batch = self.make_batch()
+        data = {
+            "batch_id": uuid.uuid4().hex,
+            "metadata": self.get_metadata(),
+            "logs": "<LOGS_JSON>",
```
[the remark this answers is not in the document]
d3dave: If we just json it we get a list of strings, not objects, because the messages are stored as json strings in the list.

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+        }
+        # batch.logs is a list of json.dump()ed strings so ",".join() it into the final json string instead of json-ing
+        # the list.
+        body = json.dumps(data).replace('"<LOGS_JSON>"', f"[{','.join(batch.logs)}]")
+        response = self.session.post(
```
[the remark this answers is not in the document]
d3dave: added.

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+            f"{self.scheme}://{self.server_address}/api/v1/logs",
+            data=body.encode("utf-8"),
+            headers={"Content-Type": "application/json"},
+            timeout=self.request_timeout,
+        )
+        return batch, response
+
+    def get_metadata(self) -> dict:
+        """Called to get metadata per batch."""
+        return {}
```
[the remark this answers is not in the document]
d3dave: Which metadata do you think should be here? The metadata is dynamic, so I expect each project to override `get_metadata` and add the fields it wants.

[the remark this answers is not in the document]
d3dave: Also, gprofiler places everything under `extra`, whereas other projects place fields directly in the metadata.

granulate_utils/glogger/messages_buffer.py
```diff
@@ -0,0 +1,64 @@
...
+        assert len(item) < self.max_total_length, "item is too long!"
+        self.buffer.append(item)
+        self.lengths.append(len(item))
+        self.total_length += len(item)
+        self.handle_overflow()
+
+    def handle_overflow(self) -> None:
+        if self.total_length >= self.max_total_length:
+            dropped = self.drop(max(1, int(self.overflow_drop_factor * self.count)))
+            logger.warning(f"Maximum total length ({self.max_total_length}) exceeded. Dropped {dropped} messages.")
```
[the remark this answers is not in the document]
d3dave: we can't report this in metadata because metadata is info that is per-log. we can report this as an additional field in the json.

tests/test_glogger.py
```diff
@@ -0,0 +1,233 @@
...
+        return f"{addr[0]}:{addr[1]}"
+
+
+def assert_buffer_attributes(handler, **kwargs):
+    mb = handler.messages_buffer
+    for k in kwargs:
+        assert getattr(mb, k) == kwargs[k]
+
+
+def assert_serial_nos_ok(serial_nos):
```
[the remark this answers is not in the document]
d3dave: - Faster to write without types
- Tests don't define an API
- If there are any type errors then most likely the tests will just fail, so type-checking the tests is redundant.

granulate_utils/glogger/__init__.py
d3dave: That's just the default. Can always be controlled using `hdlr.setLevel` if needed. DEBUG is a sane default that we use in gprofiler and elsewhere.

granulate_utils/glogger/__init__.py
d3dave: - This is called when lock is not held, so buffer is not being modified, so corruption won't happen.
- The lock is RLock (ie reentrant) so if it were held it won't cause a deadlock.

Deadlock does not worry me. The bigger fear is corruption which in this case won't happen. But I see now that it can definitely happen with the "dropped" message in the MessagesBuffer class...

rgruner (changes requested):

granulate_utils/logging/__init__.py
d3dave: Added.

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
+#
```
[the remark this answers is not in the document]
d3dave: moved to handler.py

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+    """A collection of messages sent to the server."""
+
+    logs: List[str]
+    size: int
+    head_serial_no: int
+
+
+class BatchRequestsHandler(Handler):
+    """
+    logging.Handler that accumulates logs in a buffer and flushes them periodically or when
```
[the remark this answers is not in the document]
d3dave: added readme

granulate_utils/glogger/__init__.py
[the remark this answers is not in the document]
d3dave: Reworked the entire truncation logic. Should be better now.

granulate_utils/glogger/__init__.py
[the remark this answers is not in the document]
d3dave: Reworked it to use `format` now.

granulate_utils/glogger/__init__.py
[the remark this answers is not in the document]
d3dave: Here's what the buffer looks like for example:

```python
['{"message":"test 1","serial_no":1,"severity":20,"timestamp":"2022-08-24T21:16:33.963719","logger_name":"granulate_utils.df360e179aeb8570"}',
 '{"message":"test 2","serial_no":2,"severity":20,"timestamp":"2022-08-24T21:16:33.963813","logger_name":"granulate_utils.df360e179aeb8570"}',
 '{"message":"test 3","serial_no":3,"severity":20,"timestamp":"2022-08-24T21:16:33.963887","logger_name":"granulate_utils.df360e179aeb8570"}']
```

If we json it as is we get:
```json
["{\"message\":\"test 1\",\"serial_no\":1,\"severity\":20,\"timestamp\":\"2022-08-24T21:16:33.963719\",\"logger_name\":\"granulate_utils.df360e179aeb8570\"}", "{\"message\":\"test 2\",\"serial_no\":2,\"severity\":20,\"timestamp\":\"2022-08-24T21:16:33.963813\",\"logger_name\":\"granulate_utils.df360e179aeb8570\"}", "{\"message\":\"test 3\",\"serial_no\":3,\"severity\":20,\"timestamp\":\"2022-08-24T21:16:33.963887\",\"logger_name\":\"granulate_utils.df360e179aeb8570\"}"]
```

If instead we do `f'[{",".join(buffer)}]'`:
```json
[{"message":"test 1","serial_no":1,"severity":20,"timestamp":"2022-08-24T21:16:33.963719","logger_name":"granulate_utils.df360e179aeb8570"},{"message":"test 2","serial_no":2,"severity":20,"timestamp":"2022-08-24T21:16:33.963813","logger_name":"granulate_utils.df360e179aeb8570"},{"message":"test 3","serial_no":3,"severity":20,"timestamp":"2022-08-24T21:16:33.963887","logger_name":"granulate_utils.df360e179aeb8570"}]
```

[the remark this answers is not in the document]
d3dave: The reason is for the max total length functionality. To know the length of the message we need to JSONify it. Now we can either store the original object and JSONify each message again later, or (what I have decided to do here) to store the message in it's JSON form and join all the messages together later.

Any logic performed on the message is done when it's emitted, after which it is "sealed".

granulate_utils/glogger/__init__.py
```diff
@@ -0,0 +1,189 @@
...
+            "metadata": self.get_metadata(),
+            "logs": "<LOGS_JSON>",
+        }
+        # batch.logs is a list of json.dump()ed strings so ",".join() it into the final json string instead of json-ing
+        # the list.
+        body = json.dumps(data).replace('"<LOGS_JSON>"', f"[{','.join(batch.logs)}]")
+        response = self.session.post(
+            f"{self.scheme}://{self.server_address}/api/v1/logs",
+            data=body.encode("utf-8"),
+            headers={"Content-Type": "application/json"},
```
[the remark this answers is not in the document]
d3dave: Added X-Token parameter.

granulate_utils/glogger/__init__.py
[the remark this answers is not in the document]
d3dave: The default level is 9, which is best compression. 6 is a tradeoff so we don't waste CPU.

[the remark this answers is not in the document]
d3dave: From `gzip.py`:

```python
_COMPRESS_LEVEL_FAST = 1
_COMPRESS_LEVEL_TRADEOFF = 6
_COMPRESS_LEVEL_BEST = 9
```

granulate_utils/glogger/__init__.py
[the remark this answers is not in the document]
d3dave: Added example in readme.

granulate_utils/glogger/__init__.py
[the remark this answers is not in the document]
rgruner: Then add comment?

granulate_utils/glogger/__init__.py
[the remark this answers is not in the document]
rgruner: Well we're trying to make it common for all three - let's decide on a format for how to pass the metadata and do it. I do suggest using it like gProfiler does it as it works best right now IMO.

If I'd do it over, I think we should seperate between metadata and extra - metadata will be added to each log message under metadata field and should be constant for handler (so pass it in ctor). extra can be passed for each log invocation and should be added to the log under extra.

Since we also send a lot of common metadata (instance name, instance type, etc) maybe we want to place common default values which will be used for all of our agents (cpu type, host name, etc)

granulate_utils/logging/__init__.py
d3dave: Added docs.

granulate_utils/logging/__init__.py
d3dave: renamed.

granulate_utils/glogger/__init__.py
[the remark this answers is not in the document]
d3dave: added.

granulate_utils/glogger/messages_buffer.py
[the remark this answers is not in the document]
d3dave: fixed.

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+    def emit(self, record: LogRecord) -> None:
+        # Called while lock is acquired
+        prev_dropped = self.messages_buffer.dropped
+        self.messages_buffer.append(self.format(record))
+        new_dropped = self.messages_buffer.dropped - prev_dropped
+        if new_dropped:
+            self.warn_dropped(new_dropped)
+
+    def format(self, record: LogRecord) -> str:
+        super().format(record)
```
rgruner, line 115: The super().format() has many nice features of it that will aid debugging - might be worth implementing adding with more data:
```python
formatted_message = json.loads(super().format(record))
s.update(formatted_message)
return s
``` 

like in https://github.com/logzio/logzio-python-handler/blob/master/logzio/handler.py#L89

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        formatted_timestamp = datetime.utcfromtimestamp(record.created).isoformat()
+        d = {
+            **self.get_extra_fields(record),
+            "message": record.message,
+            "serial_no": self.messages_buffer.next_serial_no,
+            "severity": record.levelno,
```
rgruner: Would suggest moving to the level name instead of the number (bit confusing and harder to query)

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+            self.truncate(d)
+            s = self.jsonify(d)
+            assert len(s) <= self.max_message_size, "did not truncate enough!"
+        return s
+
+    def get_extra_fields(self, record: LogRecord) -> dict:
+        """Override to add extra fields to formatted record."""
+        return {}
+
+    def truncate(self, d):
```
rgruner, line 137: Would change to staticmethod as it is common logic (maybe internal as well with _ as prefix)

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        minimum = 80
```
rgruner, line 138: I think the logic is a bit complex - would change to:
1. jsonify the string and understand if over the limit.
2. If over the limit understand by how much, then truncate from only the message field - that's the only thing that should do it big enough
3. Then assert that it works or drop the message (and indicate that it was dropped)

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+            "logger_name": record.name,
+        }
+        s = self.jsonify(d)
+        if len(s) > self.max_message_size:
+            self.truncate(d)
+            s = self.jsonify(d)
+            assert len(s) <= self.max_message_size, "did not truncate enough!"
+        return s
+
+    def get_extra_fields(self, record: LogRecord) -> dict:
```
rgruner, line 133: I think this fits the metadata but not the extra fields (which should be passed for each record)
See https://github.com/logzio/logzio-python-handler/blob/master/logzio/handler.py#L50 for possible implementation

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        minimum = 80
+        long_items = [(k, v) for k, v in d.items() if isinstance(v, str) and len(v) > minimum]
+        # Leave a KB buffer for all the short items + any overhead:
+        max_field_length = max(minimum, (self.max_message_size - 1024) // len(long_items))
+        for k, v in long_items:
+            d[k] = v[:max_field_length]
+        d["truncated"] = True
+
+    def warn_dropped(self, dropped: int) -> None:
+        # If we use logging, we risk getting here more than once. No can do.
```
rgruner, line 147: I really don't like this, much copied code and format will need to be changed in other places.
Would pass to server in next message as its own field (and zero the number of dropped only after we sent successfully)

This will require slight change in server.

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+                    "timestamp": datetime.utcnow().isoformat(),
+                    "logger_name": "glogger",
+                    "severity": logging.WARNING,
+                    "message": f"Maximum total length ({self.messages_buffer.max_total_length}) exceeded. "
+                    f"Dropped {dropped} messages.",
+                }
+            )
+        )
+
+    def _flush_loop(self) -> None:
```
rgruner, line 163: Would move the entire sending logic to it's own separate class - `sender`

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        """
+        :return: (batch that was sent, response)
+        """
+        # Upon every retry we will remake the batch, in case we are able to batch more messages together.
+        batch = self.make_batch()
+        batch_data = {
+            "batch_id": batch.ident,
+            "metadata": self.get_metadata(),
+            "lost": batch.lost,
+            "logs": "<LOGS_JSON>",
```
rgruner, line 213: I really don't like this, if we're trying to optimize let's add a level of cache so we won't do the json() twice.

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+            headers={
+                "Content-Encoding": "gzip",
+                "Content-Type": "application/json",
+            },
+            timeout=self.request_timeout,
+        )
+        return batch, response
+
+    def _drop_sent_batch(self, sent_batch: Batch) -> None:
+        # we don't override createLock(), so lock is not None
```
rgruner, line 231: Didn't understand the comment - what is createLock?

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+    def _send_logs(self) -> Tuple[Batch, requests.Response]:
+        """
+        :return: (batch that was sent, response)
+        """
+        # Upon every retry we will remake the batch, in case we are able to batch more messages together.
+        batch = self.make_batch()
+        batch_data = {
+            "batch_id": batch.ident,
+            "metadata": self.get_metadata(),
+            "lost": batch.lost,
```
rgruner, line 212: Can you rename to something more meaningful? Why not `dropped_logs_count`?

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+            },
+            timeout=self.request_timeout,
+        )
+        return batch, response
+
+    def _drop_sent_batch(self, sent_batch: Batch) -> None:
+        # we don't override createLock(), so lock is not None
+        with self.lock:  # type: ignore
+            # The previous lost count has been accounted by the server:
+            self.messages_buffer.dropped -= sent_batch.lost
```
rgruner, line 234: There is some discrepency in naming - dropped vs lost - let's please choose one - I suggested a naming in another comment.

glogger/handler.py
```diff
@@ -0,0 +1,277 @@
...
+                self.messages_buffer.dropped -= dropped_inadvertently
+                # Drop the remainder:
+                self.messages_buffer.drop(remaining)
+            elif remaining < 0:
+                # Account for all the messages in the batch that were considered dropped:
+                self.messages_buffer.dropped -= len(sent_batch.logs)
+                # Uh oh! We lost some messages. The dropped count now should match -remaining.
+                # It will be reported to the server at the next flush.
+
+    def get_metadata(self) -> dict:
```
rgruner, line 249: I don't think that inheritence is the nicest way to pass this - why not just pass as parameter to ctor? What are the benefits of this approach?

glogger/messages_buffer.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @property
+    def utilized(self) -> float:
+        """Total length used divided by maximum total length."""
+        return self.total_length / self.max_total_length
+
+    @property
+    def next_serial_no(self) -> int:
+        """The serial number of the next item to be inserted."""
+        return self.head_serial_no + self.count
```
rgruner, line 38: Possible race condition? If head_serial_no changes but count doesnt? Or is it atomic?

glogger/messages_buffer.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def append(self, item: str) -> None:
+        assert len(item) < self.max_total_length, "item is too long!"
```
rgruner, line 41: Thread safety here?

rgruner (changes requested):

glogger/handler.py
d3dave: Ok, then we'll have the server-side translate level name to 1-6 level number

glogger/handler.py
d3dave: The message field is actually not likely to be very large since almost all our message our short static messages. The long parts are stacktraces usually

d3dave: `exception` and `stack_trace`

glogger/handler.py
rgruner: Actually, I rethink that it makes better sense that the sdk will send the 1-6 levels, the server will accept it like that only (otherwise each SDK can send a random of something else)

glogger/handler.py
d3dave: TODO: add doc about terminology

glogger/extra.py
```diff
@@ -0,0 +1,17 @@
...
+from typing import Any, MutableMapping, Tuple
+
+
+class ExtraAdapter(logging.LoggerAdapter):
+    """
+    Logging adapter that adds an attribute named `_extra` containing a dict of all extra fields of a record.
+    This allows referring to the extras dict in a Formatter, when the extra field names differ among records.
+    """
+
+    def process(self, msg: Any, kwargs: MutableMapping[str, Any]) -> Tuple[Any, MutableMapping[str, Any]]:
```
rgruner, line 11: Would add the logic from maestro and from gProfiler that treats certain whitelist keywords differently.

rgruner, line 11: Don't see anywhere it is used.

glogger/extra.py
```diff
@@ -0,0 +1,17 @@
+import logging
```
rgruner, line 1: Rename file to extra-adapter?

glogger/extra.py
d3dave: added documentation

glogger/messages_buffer.py
```diff
@@ -22,37 +25,47 @@ def __init__(self, max_total_length: int, overflow_drop_factor: float):
...
-        self.buffer.append(item)
-        self.lengths.append(len(item))
-        self.total_length += len(item)
-        self.handle_overflow()
+        with self.lock:
+            assert len(item) < self.max_total_length, "item is too long!"
+            self.buffer.append(item)
+            self.lengths.append(len(item))
+            self.total_length += len(item)
+            self._handle_overflow_unlocked()
```
d3dave, line 54: Meant `locked`?
```suggestion
            self._handle_overflow_locked()
```

PR #64, opened 2022-08-18
Added changes to support Windows build of gprofiler

teoluwaf:
**PR Description**

The purpose of this PR is to provide support for the Windows build of gprofiler. This PR imports the pip platform package and checks if the binary is running on Windows or Linux. Depending on the platform, the major and minor version numbers of OS is retrieved in a way that doesn't break the intended use of get_kernel_release.


granulate_utils/linux/__init__.py
```diff
@@ -3,12 +3,23 @@
...
 
 def get_kernel_release() -> Tuple[int, int]:
     """Return Linux kernel version as (major, minor) tuple."""
-    major_str, minor_str = os.uname().release.split(".", maxsplit=2)[:2]
+    major_str = ""
+    minor_str = ""
+    if platform.system() == 'Linux':
+        major_str, minor_str = os.uname().release.split(".", maxsplit=2)[:2]
+    else:
+        release = platform.uname().release
```
Jongy, line 17: How does `platform.uname()` look like for Windows?

granulate_utils/linux/__init__.py
```diff
@@ -3,12 +3,23 @@
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
 import os
+import platform
 from typing import Tuple
 
 
 def get_kernel_release() -> Tuple[int, int]:
```
Jongy, line 10: I don't think this change makes sense. `get_kernel_release` is used to check kernel compatibility with features, so it makes no sense to check Windows version here. We should instead `assert platform.system() == "Linux"` and ensure this function isn't used on Windows

granulate_utils/linux/__init__.py
teoluwaf: uname_result(system='Windows', node='machine-name', release='10', version='10.0.19044', machine='AMD64')

granulate_utils/linux/__init__.py
teoluwaf: An assert won't fix the issue I ran into as the os module doesn't have a uname attribute on windows. that's why I check to see if it is running on Linux before calling os.uname(). If I don't have that check, I'll get the following error why PyInstaller attempts to load the granulate_utils/linux module:

 File "PyInstaller\loader\pyimod03_importers.py", line 495, in exec_module
  File "granulate_utils\linux\kernel_messages.py", line 30, in <module>
  File "granulate_utils\linux\__init__.py", line 15, in get_kernel_release
**AttributeError: module 'os' has no attribute 'uname'**
[25204] Failed to execute script 'pyi_build' due to unhandled exception!

granulate_utils/linux/__init__.py
Jongy: Cool

granulate_utils/linux/__init__.py
Jongy: Hmm. Logically `kernel_messags.py` should not be imported at all when we run the Windows version. Hence I see no reason to make this change. Let's remove this change and try to avoid the import in gProfiler's/PyInstaller's side?

granulate_utils/__init__.py
```diff
@@ -2,4 +2,13 @@
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
+import sys
+from functools import lru_cache
+
 __version__ = "0.0.1"
+WINDOWS_PLATFORM_NAME = "win32"
+
+@lru_cache(maxsize=None)
```
Jongy, line 11: It's okay to add it here, but we can also just add in gProfiler repo. This granulate-utils repo is for utilities shared between gProfiler and other projects, and currently `is_windows` is required only by gProfiler so we can place it there.

teoluwaf: If I move it from here, there's no way for __init__.py in gprofiler/profilers to check what modules to import. Except you are fine with me defining it in gprofiler/profilers/__init__.py before the import statements. I thought it would be cleaner to have it in your utils package.

Jongy: It can go to `gprofiler/utils/__init__.py` or possibly to `gprofiler/__init__.py` if you need it very very early in the imports flow

Jongy:
Not required - we ended up supporting Windows fully from the gProfiler repo itself.

PR #66, opened 2022-08-30
Update docker version to ~5.0.0 for pineapple compatability

uriariel:
Update docker version to ~5.0.0 for pineapple compatibility

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ psutil==5.8.0
 requests~=2.27.1
 grpcio==1.43.0
 protobuf==3.19.4
-docker==5.0.0
+docker~=5.0.0
 dataclasses==0.8; python_version < '3.7'
 typing-extensions>=4.1.0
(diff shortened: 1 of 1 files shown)


Jongy (approved):

requirements.txt
```diff
@@ -2,7 +2,7 @@ psutil==5.8.0
 requests~=2.27.1
 grpcio==1.43.0
 protobuf==3.19.4
-docker==5.0.0
+docker~=5.0.0
```
YishaiZinkin, line 5: Why not upgrade to 5.0.3 as it is in Pineapple? Or downgrade Pineapple? @Jongy

Jongy: This is a utils lib so let the users i.e gprofiler/pineapple pick the exact version

YishaiZinkin: Kk

YishaiZinkin (approved):

PR #70, opened 2022-09-11
Relax requirements

d3dave:
The requirements are too strict and prevent installation of newer patch versions.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
-psutil==5.8.0
+psutil~=5.8.0
 requests~=2.27.1
-grpcio==1.43.0
-protobuf==3.19.4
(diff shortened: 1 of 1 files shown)


Jongy (approved):

requirements.txt
```diff
@@ -1,9 +1,9 @@
-psutil==5.8.0
+psutil~=5.8.0
 requests~=2.27.1
-grpcio==1.43.0
-protobuf==3.19.4
+grpcio~=1.43.0
```
michelhe, line 3: Mmm maybe should have left grpcio pinned as its build time is substantial?

d3dave: Users responsible for pinning if needed.

PR #73, opened 2022-09-14
Added golang and node process utils

grariel:
Most of the code is copied from:
https://github.com/Granulate/gprofiler/blob/master/gprofiler/profilers/perf.py#L272

And will be used there as well.

### DIFF
--- a/granulate_utils/golang.py
+++ b/granulate_utils/golang.py
@@ -1,0 +1,41 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+import functools
+import struct
+from typing import Optional
+
+from psutil import NoSuchProcess, Process
+
+from granulate_utils.linux.elf import read_elf_symbol, read_elf_va
+
+
+def is_golang_process(process: Process) -> bool:
+    return get_process_golang_version(process) is not None
+
+
+@functools.lru_cache(maxsize=4096)
+def get_process_golang_version(process: Process) -> Optional[str]:
+    elf_path = f"/proc/{process.pid}/exe"
+    try:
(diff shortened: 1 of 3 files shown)


michelhe:
put these in a their own module, don't throw it in the package root

grariel:
> put these in a their own module, don't throw it in the package root

Python and Java are also in the root folder.

granulate_utils/node.py
```diff
@@ -0,0 +1,6 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+
+DETECTED_NODE_PROCESSES_REGEX = r"^.+/libnode\.so"
```
Jongy, line 6: There is no such thing as `libnode`. Node process identification is done by `/proc/pid/exe`.

grariel: Apparently it's only loaded on some versions. Changed to gProfiler's method, so now all the code is copied.

michelhe:
See that the runtime detection code in gProfiler has gone through some changes here :
https://github.com/Granulate/gprofiler/pull/460

granulate_utils/golang.py
```diff
@@ -0,0 +1,42 @@
...
+from granulate_utils.linux.elf import read_elf_symbol, read_elf_va
+
+
+def is_golang_process(process: Process) -> bool:
+    return get_process_golang_version(process) is not None
+
+
+@functools.lru_cache(maxsize=4096)
+def get_process_golang_version(process: Process) -> Optional[str]:
+    elf_path = f"/proc/{process.pid}/exe"
```
michelhe, line 21: use `granulate_utils.linux.process.process_exe`

michelhe (approved):

PR #74, opened 2022-09-14
linux: process: Add read_proc_file helper

d3dave:
Add `read_proc_file` to read arbitrary `/proc/PID/...` file, while translating all the needed exceptions appropriately.

I have verified that the changed methods still work fine.

### DIFF
--- a/granulate_utils/linux/process.py
+++ b/granulate_utils/linux/process.py
@@ -102,6 +102,9 @@ def _read_process_memory(process: psutil.Process, addr: int, size: int) -> bytes
 def _translate_errors(process: psutil.Process) -> Generator[None, None, None]:
     try:
         yield
+        # Don't use the result if PID has been reused
+        if not process.is_running():
+            raise psutil.NoSuchProcess(process.pid)
     except PermissionError:
         raise psutil.AccessDenied(process.pid)
     except ProcessLookupError:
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/process.py
```diff
@@ -82,16 +92,21 @@ def _read_process_auxv(process: psutil.Process, auxv_id: int) -> int:
...
-        raise psutil.NoSuchProcess(process.pid)
 
 
-def read_process_execfn(process: psutil.Process) -> str:
-    # reads process AT_EXECFN
-    addr = _read_process_auxv(process, AT_EXECFN)
-    fn = _read_process_memory(process, addr, PATH_MAX)
-    return fn[: fn.index(b"\0")].decode()
+@contextmanager
+def _translate_errors(process: psutil.Process):
```
michelhe, line 102: Add return type hint (Genetator[None,None,None])

michelhe (changes requested):

granulate_utils/linux/process.py
```diff
@@ -63,12 +64,21 @@ def get_mapped_dso_elf_id(process: psutil.Process, dso_part: str) -> Optional[st
         return None
 
 
+def read_proc_file(process: psutil.Process, name: str) -> bytes:
```
Jongy, line 67: gProfiler's verifies that `process` is still running afterwards (protects from PID reuse). Let's add this verification here.

Also - please open the relevant PR in gProfiler so it uses your new API, to avoid keeping 2 copies of the function.

granulate_utils/linux/process.py
Jongy: Done

michelhe (approved):
LGTM

Jongy (approved):
LGTM, please apply the change in gProfiler

PR #75, opened 2022-09-20
cgroups: memory: fixed set_limit_in_bytes

marina-vrublevsky:
1. Ensure we reset `memsw.limit_in_bytes` (in case it exists) in order to allow changing `limit_in_bytes` in case it's smaller than `memsw.limit_in_bytes` (or we will experience `write error: Invalid argument`)
2. Fix `reset_memory_limit` which previously didn't reset  `memsw.limit_in_bytes` (in case it exists)

### DIFF
--- a/granulate_utils/linux/cgroups/memory_cgroup.py
+++ b/granulate_utils/linux/cgroups/memory_cgroup.py
@@ -27,12 +27,13 @@ class MemoryCgroup(BaseCgroup):
             pass
 
     def set_limit_in_bytes(self, limit: int) -> None:
-        # in case memsw_limit_in_bytes file exists we need to reset it in order to
-        # change limit_in_bytes in case it's smaller than memsw_limit_in_bytes
+        # in case memsw.limit_in_bytes file exists we need to reset it in order to
+        # change limit_in_bytes in case it's smaller than memsw.limit_in_bytes
         self._set_memsw_limit_in_bytes(-1)
         self.write_to_control_file(self.limit_in_bytes, str(limit))
+
+        # memsw.limit_in_bytes is already set to -1
         if limit != -1:
-            # memsw_limit_in_bytes already happend for -1
             self._set_memsw_limit_in_bytes(limit)
 
     def reset_memory_limit(self) -> None:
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/cgroups/memory_cgroup.py
```diff
@@ -18,14 +18,22 @@ def get_memory_limit(self) -> int:
...
             pass
 
+    def set_limit_in_bytes(self, limit: int) -> None:
+        # in case memsw_limit_in_bytes file exists we need to reset it in order to
+        # change limit_in_bytes in case it's smaller than memsw_limit_in_bytes
+        self._set_memsw_limit_in_bytes(-1)
+        self.write_to_control_file(self.limit_in_bytes, str(limit))
+        if limit != -1:
+            # memsw_limit_in_bytes already happend for -1
+            self._set_memsw_limit_in_bytes(limit)
```
d3dave, line 36:
```suggestion
        # in case memsw.limit_in_bytes file exists we need to reset it in order to
        # change limit_in_bytes in case it's smaller than memsw.limit_in_bytes
        self._set_memsw_limit_in_bytes(-1)
        self.write_to_control_file(self.limit_in_bytes, str(limit))
        # memsw.limit_in_bytes already set to -1
        if limit != -1:
            self._set_memsw_limit_in_bytes(limit)
```

granulate_utils/linux/cgroups/memory_cgroup.py
```diff
@@ -18,14 +18,23 @@ def get_memory_limit(self) -> int:
     def get_max_usage_in_bytes(self) -> int:
         return int(self.read_from_control_file(self.max_usage_in_bytes))
 
-    def set_limit_in_bytes(self, limit: int) -> None:
-        self.write_to_control_file(self.limit_in_bytes, str(limit))
+    def _set_memsw_limit_in_bytes(self, limit: int) -> None:
         try:
             self.write_to_control_file(self.memsw_limit_in_bytes, str(limit))
         except PermissionError:
```
motisaroka, line 24: Any other exception should be raised here?

d3dave (approved):

granulate_utils/linux/cgroups/memory_cgroup.py
marina-vrublevsky: PermissionError means the file doesn't exist and this is the case we are handling here - imo if it's another exception we need to raise on it in order not to lose it. @d3dave wdyt?

marina-vrublevsky: + I place a comment just under it explaining the reason

PR #77, opened 2022-09-21
Added get_elf_arch() that reads the binary architecture from the ELF section of the file

tomersa:
Added get_elf_arch() that reads the binary architecture from the ELF section of the file

# Tests
- [x] - Run the function on an arm binary and the function returned - 'AArch64'
- [x] - Run the function on an x86_64 binary and the function returned - 'x64'

### DIFF
--- a/granulate_utils/linux/elf.py
+++ b/granulate_utils/linux/elf.py
@@ -12,7 +12,7 @@ from elftools.elf.sections import NoteSection  # type: ignore
 
 def get_elf_arch(path: str) -> str:
     """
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/elf.py
```diff
@@ -10,6 +10,16 @@
 from elftools.elf.sections import NoteSection  # type: ignore
 
 
+def get_elf_arch(path: str) -> str:
+    """
+
```
kmaork, line 15: remove empty line?

kmaork (approved):

PR #82, opened 2022-10-03
Added ELFERROR

tomersa:


requirements.txt
```diff
@@ -7,3 +7,4 @@ dataclasses~=0.8; python_version < '3.7'
 typing-extensions>=4.1.0
 pyelftools~=0.28
 packaging~=21.2
+importlib_metadata~=4.8
```
tomersa, line 10: That was added in order to pass the tests (They were broken on master because of this import). See comment below from the commit for more details:
```
    * Tests fail on python3.7 because of a problem with the importlib_metadata package.
    The issue was fixed using the workaround suggested here: https://github.com/python/importlib_metadata/issues/406#issuecomment-1264682770

```

DanielShaulov (approved):

requirements.txt
d3dave: Well the issue is actually with flake8, so I wouldn't put this in the requirements.txt, I'd put a pip install for this in the CI workflow.

tomersa: Here you go: https://github.com/Granulate/granulate-utils/pull/84/files

PR #84, opened 2022-10-04
Moving requirement to handle flake8 issue to the CI instead of the requirements.txt

tomersa:


.github/workflows/glogger.yml
```diff
@@ -24,6 +24,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements.txt
         pip install -r dev-requirements.txt
+        pip install --force-reinstall importlib_metadata~=4.8  # Added requirement separately since it stems from an errorneous lint. See the following link for more details: https://github.com/Granulate/granulate-utils/pull/82#discussion_r985745999
```
Jongy, line 27: So why separately and not in dev-requirements? And why --force-reinstall?

tomersa: Maybe the comment wasn't clear, `importlib_metadata==5` is installed as it is a requirement of 1 of the packages used by granulate-utils/gLogger. I need importlib_metadata~=4.8 so I must reinstall it in order for it to work. 

It is separate from dev-requirements.yml because it should be temporary until they fix the problematic requirements in the link suggested.

Jongy: I managed. Do the same fix here and remove the `importlib_metadata` thing

tomersa:
already fixed

Jongy:
> already fixed

@tomersa - this is not fixed, you kept the `importlib_metadata` in `requirements.txt` instead of upgrading flake8 as [we did in gProfiler](https://github.com/Granulate/gprofiler/pull/519). Please revert the change done [here](https://github.com/Granulate/granulate-utils/pull/82/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552) and instead upgrade flake8

tomersa:
I've checked it using a temp PR and all the CI tests passed:
https://github.com/Granulate/granulate-utils/pull/90/checks

Jongy:
Fixed in https://github.com/Granulate/granulate-utils/pull/92

PR #85, opened 2022-10-06
elf: Add get_libc_type

DanielShaulov:
Adds `get_libc_type` function to `elf.py`, it distinguishes between dynamic glibc/musl and for static - between libc existing an not existing.
It is harder to distinguish between musl/glibc in static binaries and technically, not really needed for our use cases.

I am 50/50 about splitting it to `get_dynamic_libc_type` and `is_static_libc` that just do their respective parts, lmk what you think.

As an aside - we really need to move this to a class so we don't need to re-read the elf for each function. We can keep the global functions as a convenience wrapper. Specifically - `get_libc_type` reads the elf twice, lmk if you want me to do the refactor to a class right now to avoid this.

### DIFF
--- a/granulate_utils/linux/elf.py
+++ b/granulate_utils/linux/elf.py
@@ -85,8 +85,10 @@ def read_elf_symbol(path: str, sym_name: str, size: int) -> Optional[bytes]:
         if symtab is None:
             return None
         symbols = symtab.get_symbol_by_name(sym_name)
-        if symbols is None or len(symbols) != 1:
+        if symbols is None:
             return None
+        if len(symbols) != 1:
+            raise Exception(f"Multiple symbols match the same name: {sym_name!r}")
         return _read_va_from_elf(elf, symbols[0].entry.st_value, size)
 
 
@@ -106,41 +108,43 @@ def get_symbol_addr(path: str, sym_name: str) -> Optional[int]:
         if symtab is None:
             return None
         symbols = symtab.get_symbol_by_name(sym_name)
-        if symbols is None or len(symbols) != 1:
+        if symbols is None:
             return None
+        if len(symbols) != 1:
+            raise Exception(f"Multiple symbols match the same name: {sym_name!r}")
         return symbols[0].entry.st_value
 
 
-def get_dt_needed(path: str) -> List[str]:
-    dt_needed: List[str] = []
+def get_dt_needed(path: str) -> Optional[List[str]]:
     with open(path, "rb") as f:
         elf = ELFFile(f)
         dynamic_section = elf.get_section_by_name(".dynamic")
         if dynamic_section is None:
-            return dt_needed
-        for tag in dynamic_section.iter_tags():
-            if tag.entry.d_tag == "DT_NEEDED":
-                dt_needed.append(tag.needed)
-    return dt_needed
+            return None
+        return [tag.needed for tag in dynamic_section.iter_tags() if tag.entry.d_tag == "DT_NEEDED"]
 
 
 def get_libc_type(path: str) -> LibcType:
     dt_needed = get_dt_needed(path)
-    found_glibc = False
-    found_musl = False
-    for needed in dt_needed:
-        if "libc.so.6" in needed:
-            found_glibc = True
-        if "libc.musl" in needed:
-            found_musl = True
-        if "ld-musl" in needed:
-            found_musl = True
-    if found_glibc and found_musl:
-        raise Exception("Found both musl and glibc in the same binary")
-    if found_musl:
-        return LibcType.DYNAMIC_MUSL
-    if found_glibc:
-        return LibcType.DYNAMIC_GLIBC
+    if dt_needed is not None:
+        found_glibc = False
+        found_musl = False
+        for needed in dt_needed:
+            if "libc.so.6" in needed:
+                found_glibc = True
+            if "ld-linux" in needed:
+                found_glibc = True
+            if "libc.musl" in needed:
+                found_musl = True
+            if "ld-musl" in needed:
+                found_musl = True
+        if found_glibc and found_musl:
+            raise Exception(f"Found both musl and glibc in the same binary: {path!r}")
+        if found_musl:
+            return LibcType.DYNAMIC_MUSL
+        if found_glibc:
+            return LibcType.DYNAMIC_GLIBC
+        raise Exception(f"Foind a dynamic binary without a libc: {path!r}")
 
     # This symbol exists in both musl and glibc, and is even used by musl to recognize a DSO as a libc.
     # They even comment that it works on both musl and glibc.
--- a/granulate_utils/linux/elf.py
+++ b/granulate_utils/linux/elf.py
@@ -10,7 +10,6 @@ from typing import List, Optional, cast
 from elftools.elf.elffile import ELFError, ELFFile  # type: ignore
 from elftools.elf.sections import NoteSection  # type: ignore
 
-
 __all__ = ["ELFError"]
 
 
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/elf.py
```diff
@@ -90,3 +97,52 @@ def is_statically_linked(path: str) -> bool:
...
+        if "ld-musl" in needed:
+            found_musl = True
+    if found_glibc and found_musl:
+        raise Exception("Found both musl and glibc in the same binary")
+    if found_musl:
+        return LibcType.DYNAMIC_MUSL
+    if found_glibc:
+        return LibcType.DYNAMIC_GLIBC
+
+    if get_symbol_addr(path, "__libc_start_main") is not None:
```
michelhe, line 145: Is this true for both glibc and musl?

DanielShaulov: Yep - musl even uses this to discover if you are trying to dlopen another libc to redirect you to dlopen libc.so, with a comment stating this should catch both glibc and musl, so I am pretty confident:

image

granulate_utils/linux/elf.py
```diff
@@ -90,3 +97,52 @@ def is_statically_linked(path: str) -> bool:
...
+    dt_needed: List[str] = []
+    with open(path, "rb") as f:
+        elf = ELFFile(f)
+        dynamic_section = elf.get_section_by_name(".dynamic")
+        if dynamic_section is None:
+            return dt_needed
+        for tag in dynamic_section.iter_tags():
+            if tag.entry.d_tag == "DT_NEEDED":
+                dt_needed.append(tag.needed)
+    return dt_needed
```
Jongy, line 124:
```suggestion
        return [tag.needed for tag in dynamic_section.iter_tags() if tag.entry.d_tag == "DT_NEEDED"]
```

granulate_utils/linux/elf.py
```diff
@@ -90,3 +97,52 @@ def is_statically_linked(path: str) -> bool:
...
+
+
+def get_libc_type(path: str) -> LibcType:
+    dt_needed = get_dt_needed(path)
+    found_glibc = False
+    found_musl = False
+    for needed in dt_needed:
+        if "libc.so.6" in needed:
```
Jongy, line 132: What about `ld-linux`?

granulate_utils/linux/elf.py
```diff
@@ -90,3 +97,52 @@ def is_statically_linked(path: str) -> bool:
...
+            found_glibc = True
+        if "libc.musl" in needed:
+            found_musl = True
+        if "ld-musl" in needed:
+            found_musl = True
+    if found_glibc and found_musl:
+        raise Exception("Found both musl and glibc in the same binary")
```
Jongy, line 139:
```suggestion
        raise Exception(f"Found both musl and glibc in the same binary: {path!r}")
```

granulate_utils/linux/elf.py
Jongy: Cool trick. Document it here, permalink to this comment would do.

Also - seems like this logic checks for both syms, not just `__libc_start_main`, any reason you're not following it?

granulate_utils/linux/elf.py
```diff
@@ -90,3 +97,52 @@ def is_statically_linked(path: str) -> bool:
...
+
+
+def get_symbol_addr(path: str, sym_name: str) -> Optional[int]:
+    with open(path, "rb") as f:
+        elf = ELFFile(f)
+        symtab = elf.get_section_by_name(".symtab")
+        if symtab is None:
+            return None
+        symbols = symtab.get_symbol_by_name(sym_name)
+        if symbols is None or len(symbols) != 1:
```
Jongy, line 109: `len(symbols) != 1` should be considered an error state in this function, I think. Let's raise on it?

granulate_utils/linux/elf.py
Jongy: Also, code checks for != 0 and you check for "existing", but I guess that's enough as well

Jongy:
> I am 50/50 about splitting it to get_dynamic_libc_type and is_static_libc that just do their respective parts, lmk what you think.

I like it as one function.

> As an aside - we really need to move this to a class so we don't need to re-read the elf for each function. We can keep the global functions as a convenience wrapper. Specifically - get_libc_type reads the elf twice, lmk if you want me to do the refactor to a class right now to avoid this.

I'm not sure a class is needed - maybe accepting a `Union[ELFFile, str]` is enough? The outer function can then pass its own `ELFFile` to both.

granulate_utils/linux/elf.py
DanielShaulov: > Also - seems like this logic checks for both syms, not just __libc_start_main, any reason you're not following it?

I am using this image as a proof that "glibc and musl always have __libc_start_main", not as "this is the way to check".
They are using two symbols to decrease the chance for a false positive and loading the wrong thing. For use, we have a static binary that we just want to know if it uses a libc or if it is "bare" like staic go binaries, so just that symbol is enough to say "it has a libc", we would rather make the mistake in the other direction.

I can add a validation that `stdin` exists and raise an exception if it doesn't - if you feel it is safer, but returning NO_LIBC after finding `__libc_start_main` just because `stdin` is not there just sounds plain wrong.

DanielShaulov: Adding a descriptive comment and link to musl code

granulate_utils/linux/elf.py
DanielShaulov: You are absolutely correct, I copied it from `read_elf_symbol` (will be refactored to use same code after I add class or accept `Union`) - and I think it should be an error there too.

granulate_utils/linux/elf.py
DanielShaulov: added

granulate_utils/linux/elf.py
DanielShaulov: Also changed to return `None` if no dynamic section instead of empty list

DanielShaulov:
> I'm not sure a class is needed - maybe accepting a Union[ELFFile, str] is enough? The outer function can then pass its own ELFFile to both.

umm.. python is a bit shitty with conditional context managers. We want to do "with open(path)" if we got a path, but just use the file if we got a file, and that just can't be done without moving all the code inside the `with` to a function or duplicating it in the `if` branches. The only reasonable way is to have a context manager that accepts both - and if it is already and ELFFile, return it and do nothing on exit, and if it is not then open and close.
At this point a class is better IMO

DanielShaulov:
OK - I did the contextmanager that accepts everything and yields an ELFFile, I am fine with it, looks better than a class

granulate_utils/linux/elf.py
```diff
@@ -40,53 +59,97 @@ def get_elf_buildid(path: str) -> Optional[str]:
...
+                    found_musl = True
+                if "ld-musl" in needed:
+                    found_musl = True
+            if found_glibc and found_musl:
+                raise Exception(f"Found both musl and glibc in the same binary: {elf.stream.name!r}")
+            if found_musl:
+                return LibcType.DYNAMIC_MUSL
+            if found_glibc:
+                return LibcType.DYNAMIC_GLIBC
+            raise Exception(f"Foind a dynamic binary without a libc: {elf.stream.name!r}")
```
michelhe, line 147:
```suggestion
            raise Exception(f"Found a dynamic binary without a libc: {elf.stream.name!r}")
```

granulate_utils/linux/elf.py
Jongy: > I can add a validation that stdin exists and raise an exception if it doesn't - if you feel it is safer, but returning NO_LIBC after finding __libc_start_main just because stdin is not there just sounds plain wrong.

Sounds okay.

I am resolving this comment

Jongy (approved):

PR #87, opened 2022-10-07
Add wrapper for catching FileNotFound Exception

mpozniak95:
It wraps function get_elf_id(), if that function throws FileNotFound exception then wrapper checks if not found file is /proc/{pid}/*, if it is, wrapper changes exception to NoSuchProcess.

Connected to: https://github.com/Granulate/gprofiler/issues/503

### DIFF
--- a/granulate_utils/linux/elf.py
+++ b/granulate_utils/linux/elf.py
@@ -30,20 +30,14 @@ def raise_nosuchprocess(func: Callable[P, R]) -> Callable[P, R]:
                 if e.filename.split("/")[2].isalnum():
                     # Take pid from /proc/{pid}/*
                     pid = int(e.filename.split("/")[2])
+                    # Check if number from /proc/{pid} is actually a pid number
                     with open("/proc/sys/kernel/pid_max") as pid_max_file:
                         pid_max = int(pid_max_file.read())
                     if pid <= pid_max:
                         # Check if pid is running
-                        if psutil.pid_exists(pid):
-                            raise e
-                        else:
+                        if not psutil.pid_exists(pid):
                             raise psutil.NoSuchProcess(pid)
-                    else:
-                        raise e
-                else:
-                    raise e
-            else:
-                raise e
+            raise e
     return inner
 
 
--- a/granulate_utils/linux/elf.py
+++ b/granulate_utils/linux/elf.py
@@ -37,6 +37,7 @@ def raise_nosuchprocess(func: Callable[P, R]) -> Callable[P, R]:
                         if not psutil.pid_exists(pid):
                             raise psutil.NoSuchProcess(pid)
             raise e
+
     return inner
 
 
--- a/granulate_utils/linux/elf.py
+++ b/granulate_utils/linux/elf.py
@@ -18,7 +18,7 @@ def catch_filenotfound(func):
         try:
             func(*args, **kwargs)
         except FileNotFoundError as e:
-            if ("/proc/" in e.filename):
+            if (e.filename.startswith("/proc/")):
                 # Take pid from /proc/{pid}/*
                 pid = e.filename.split("/")[2]
                 raise psutil.NoSuchProcess(pid)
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/elf.py
```diff
@@ -12,6 +13,20 @@
 __all__ = ["ELFError"]
 
 
+def catch_filenotfound(func):
```
Jongy, line 16: Nice approach! (decorator).

A few suggestions:
1. I see that this will also fix the problem for `get_mapped_dso_elf_id` - good.
2. I suggest a more descriptive name - maybe `raise_nosuchprocess`? That's more important than the `catch filenotfound` part.
3. This is not robust enough. For example, it could raise `NoSuchProcess` if I try `get_elf_id("/proc/blablabla")` (where blablabla is not even a PID). It could also raise `NoSuchProcess` if I try `/proc/pid/root/path/doesnt/exist`, even if `pid` is running indeed. I suggest - in the decorator you check if the path starts with `/proc/`, and if so, is the 2nd component a number, and if so, is it running, and if not - raise.

Jongy, line 16: Make sure we don't mess up the typing of the wrapped function. You can follow this guide: https://rednafi.github.io/reflections/static-typing-python-decorators.html, it looks okay

Jongy (changes requested):
Done reviewing :)

granulate_utils/linux/elf.py
mpozniak95: 1. Done
2. Done
3. Done. Also added check if number from /proc/{number} is a number from PID number range. This avoids getting OverflowError from `psutil.pid_exists(pid)` for big numbers.

mpozniak95: Done

Jongy: The max PID check might be a bit of an overkill, but it's indeed as robust as we'll go. So okay :)

granulate_utils/linux/elf.py
```diff
@@ -4,13 +4,41 @@
...
+            if e.filename.startswith("/proc/"):
+                if e.filename.split("/")[2].isalnum():
+                    # Take pid from /proc/{pid}/*
+                    pid = int(e.filename.split("/")[2])
+                    # Check if number from /proc/{pid} is actually a pid number
+                    with open("/proc/sys/kernel/pid_max") as pid_max_file:
+                        pid_max = int(pid_max_file.read())
+                    if pid <= pid_max:
+                        # Check if pid is running
+                        if not psutil.pid_exists(pid):
```
Jongy, line 37: It is true though, that `pid_exists` is a better check than what we have in `read_proc_file` - we use `os.path.exists("/proc/pid")`, but this will return True also if `pid` is a `tid` of another process (e.g due to reuse). `pid_exists` uses different means to check, to avoid this confusion. Note @d3dave .

Jongy (changes requested):
1 more question, and fix linters :)

Jongy:
Btw, did you verify that mypy accepts the typing tricks for the decorator? (i.e that `get_elf_id` still has the typing specs as it had)

granulate_utils/linux/elf.py
d3dave: Also, instead of `exists(/proc/pid)` should have just `kill(pid, 0)`

granulate_utils/linux/elf.py
mpozniak95: No, it wasn't for an actual PID. I was just wondering what will happen if I try to get some really huge number instead of real PID (f.e. `get_elf_id("/proc/3276842132132323/exe")` -> resulted with OverflowError for `psutil.pid_exists(pid)`).

mpozniak95:
```
mypy granulate_utils/linux/elf.py
Success: no issues found in 1 source file
```

granulate_utils/linux/elf.py
Jongy: `pid_exists` does extra checks because they claim `os.kill` is bad as well :dumb:

Jongy:
Looks good!

Jongy (approved):

PR #89, opened 2022-10-10
Use read_proc_file in get_cgroups()

d3dave:
1. Updates `get_cgroups()` to use `read_proc_file` to catch errors better.
2. Removes ability to specify plain int pid to `get_process_container_id`, to force users to rely on `psutil.Process` at all times. (To reap the advantages of `psutil.Process` it must be used at all times, as early as possible.)
3. Makes `get_cgroups()` default to getting the cgroups of the current process if the parameter is not given. This is a shorthand for `get_cgroups(os.getpid())` which appears a couple times in the code.

### DIFF
--- a/granulate_utils/linux/cgroups/cgroup.py
+++ b/granulate_utils/linux/cgroups/cgroup.py
@@ -3,13 +3,13 @@
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
 
-from pathlib import Path
 from typing import List, Mapping, Optional, Tuple
 
-from psutil import NoSuchProcess
+import psutil
 
 from granulate_utils.linux import ns
 from granulate_utils.linux.mountinfo import iter_mountinfo
+from granulate_utils.linux.process import read_proc_file
 
 SUBSYSTEMS = {
     "blkio",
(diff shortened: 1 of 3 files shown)


Jongy (approved):

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -28,21 +28,19 @@
 }
 
 
-def get_cgroups(pid: int) -> List[Tuple[str, List[str], str]]:
+def get_cgroups(process: Optional[psutil.Process] = None) -> List[Tuple[str, List[str], str]]:
```
michelhe, line 31: You broke the API, please next time do `Union[int, Process]`

d3dave: Fixed, not broke. You need a Process object in this implementation for proper error handling

d3dave: Did you read the PR description?

d3dave: What would you rely on? Runtime failures? Documentation? 

Runtime failures would happen anyway, so the reliance is not strictly on static type checking. And you can't rely on documentation because it needs to be read by people. You can't rely on versioning either, because it doesn't tell you what and where to fix.

d3dave: The only alternative would be to leave everything broken and create a new API, which defeats the purpose of the change.

Jongy: I agree with @d3dave here. We rely on mypy here. This is an improvement over the previous api.

PR #92, opened 2022-10-12
Upgrade flake8

Jongy:
Follows https://github.com/Granulate/gprofiler/pull/519

Fixes what was added in #82 and I wanted fixed in https://github.com/Granulate/granulate-utils/pull/84


tomersa (approved):
LGTM

PR #95, opened 2022-10-23
Added is_kernel_process function and used it in golang and node runtimes

grariel:
The purpose of this PR is to avoid checking if kernel threads are golang/node processes.

### DIFF
--- a/granulate_utils/golang.py
+++ b/granulate_utils/golang.py
@@ -14,7 +14,7 @@ from granulate_utils.linux.process import is_kernel_thread
 
 
 def is_golang_process(process: Process) -> bool:
-    return (not is_kernel_thread(process)) and (get_process_golang_version(process) is not None)
+    return not is_kernel_thread(process) and get_process_golang_version(process) is not None
 
 
 @functools.lru_cache(maxsize=4096)
--- a/granulate_utils/golang.py
+++ b/granulate_utils/golang.py
@@ -10,10 +10,11 @@ from typing import Optional
 from psutil import NoSuchProcess, Process
 
 from granulate_utils.linux.elf import read_elf_symbol, read_elf_va
+from granulate_utils.linux.process import is_kernel_thread
 
 
 def is_golang_process(process: Process) -> bool:
-    return get_process_golang_version(process) is not None
+    return (not is_kernel_thread(process)) and (get_process_golang_version(process) is not None)
 
 
 @functools.lru_cache(maxsize=4096)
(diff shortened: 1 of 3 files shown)


granulate_utils/golang.py
```diff
@@ -10,10 +10,11 @@
 from psutil import NoSuchProcess, Process
 
 from granulate_utils.linux.elf import read_elf_symbol, read_elf_va
+from granulate_utils.linux.process import is_kernel_thread
 
 
 def is_golang_process(process: Process) -> bool:
-    return get_process_golang_version(process) is not None
+    return (not is_kernel_thread(process)) and (get_process_golang_version(process) is not None)
```
Jongy, line 17:
```suggestion
    return (not is_kernel_thread(process) and get_process_golang_version(process) is not None
```

Not pythonic

granulate_utils/linux/process.py
```diff
@@ -128,3 +128,7 @@ def is_process_basename_matching(process: psutil.Process, basename_pattern: str)
         return True
 
     return False
+
+
+def is_kernel_thread(process: psutil.Process) -> bool:
```
Jongy, line 133: This is the logic from `pgrep_exe` right?

granulate_utils/node.py
```diff
@@ -5,8 +5,8 @@
 
 from psutil import Process
 
-from granulate_utils.linux.process import is_process_basename_matching
+from granulate_utils.linux.process import is_process_basename_matching, is_kernel_thread
 
 
 def is_node_process(process: Process) -> bool:
-    return is_process_basename_matching(process, r"^node$")
+    return (not is_kernel_thread(process)) and is_process_basename_matching(process, r"^node$")
```
Jongy, line 12: Same comments on the `( )`

granulate_utils/linux/process.py
Jongy: Make sure to update `pgrep_exe` to use it

grariel: > This is the logic from `pgrep_exe` right?

Yes.

> Make sure to update `pgrep_exe` to use it

Already did:
https://github.com/Granulate/gprofiler/pull/554

d3dave (commented):
I think it's nicer if we have `get_process_runtime()` which will return node/go/python/java/etc. which does all the checks.

d3dave:
> I think it's nicer if we have `get_process_runtime()` which will return node/go/python/java/etc. which does all the checks.

Can even subclass psutil.Process to RuntimeProcess or smth like that which caches it in a class member.

grariel:
> > I think it's nicer if we have `get_process_runtime()` which will return node/go/python/java/etc. which does all the checks.
> 
> Can even subclass psutil.Process to RuntimeProcess or smth like that which caches it in a class member.

Nice, opened a ticket for it:
https://app.clickup.com/t/3035nrz

Jongy (approved):

Jongy (approved):

PR #96, opened 2022-11-02
Added cache to get_process_cotainer_id

grariel:
Possible implications: increased memory usage, and duplicate cache for `get_container_name` implementations.


Jongy:
Napkin math - if a `Process` object is 2k then this cache reaches 8mb and we haven't proven this function to be heavy, so currently we'll hold it

PR #97, opened 2022-11-08
ns: run_in_ns(): Remove passthrough_exception

Jongy:
It's a bad practice to have it defaulting to False, and it's a bad practice to use it with False anyway

### DIFF
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -205,7 +205,6 @@ def run_in_ns(
     nstypes: List[str],
     callback: Callable[[], T],
     target_pid: int = 1,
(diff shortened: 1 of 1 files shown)


michelhe (approved):

michelhe (changes requested):
I prefer that we keep it default `True` because otherwise when we upgrade the utils in our profiler we will have to remove the argument from every callsite

michelhe (approved):
But you know what the CI will fail anyway if we bump the `granulate-utils` version refactoring  the callsites so I guess its okay to remove..

d3dave:
Finally. Thanks

PR #98, opened 2022-11-09
Upgrade dev dependencies

Jongy:


Jongy:
We need to get rid of the Python 3.6 usage, it doesn't receive upgrades of packages (no mypy beyond 0.971 it seems). https://github.com/Granulate/gprofiler/issues/583

Meanwhile, I think i'll downgrade to 0.971

d3dave (approved):

PR #101, opened 2022-11-13
Added more parsing to cgroup classes

grariel:
Added:

- Ability to get memory usage
- `cpuacct` class that allows getting cpu usage
- `pids_in_cgroup` propery.

### DIFF
--- a/granulate_utils/linux/cgroups/base_cgroup.py
+++ b/granulate_utils/linux/cgroups/base_cgroup.py
@@ -12,6 +12,7 @@ from granulate_utils.linux.cgroups.cgroup import SUBSYSTEMS, find_v1_hierarchies
 
 class BaseCgroup:
     predefined_cgroups = ["kubepods", "docker", "ecs"]
+    cgroup_procs = 'cgroup.procs'
     _v1_hierarchies: Optional[Mapping[str, str]] = None
 
     def __init__(self) -> None:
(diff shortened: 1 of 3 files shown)


grariel:
Don't CR yet, decided to add more features to this PR.

grariel:
@DanielShaulov ready for CR.

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -55,6 +56,10 @@ def cgroup_mount_path(self) -> Path:
     def cgroup_path(self) -> Path:
         return Path(self.get_cgroup_hierarchies()[self.subsystem])
 
+    @property
```
DanielShaulov, line 59: This is a heavy operation and is not cached - should be a function IMO

granulate_utils/linux/cgroups/base_cgroup.py
```diff
@@ -55,6 +56,10 @@ def cgroup_mount_path(self) -> Path:
...
+    def pids_in_cgroup(self) -> List[int]:
+        return [int(proc) for proc in self.read_from_control_file(self.cgroup_procs).split()]
```
DanielShaulov, line 61: Return a `Set` instead to remove dups?

image

DanielShaulov (changes requested):
Small comments

granulate_utils/linux/cgroups/base_cgroup.py
grariel: Great catch.

granulate_utils/linux/cgroups/base_cgroup.py
grariel: Indeed, from PEP8:
> Avoid using properties for computationally expensive operations; the attribute notation makes the caller believe that access is (relatively) cheap.

Fixed.

DanielShaulov (approved):

PR #103, opened 2022-11-23
Add utility function that returns a cgroup controller folder

IlayRosenberg:
Implemented and exported `get_cgroup_mount(controller)`

### DIFF
--- a/granulate_utils/linux/cgroups/cgroup.py
+++ b/granulate_utils/linux/cgroups/cgroup.py
@@ -96,5 +96,5 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
     return None
 
 
-def is_valid_controller(controller: str) -> bool:
+def is_known_controller(controller: str) -> bool:
     return controller in SUBSYSTEMS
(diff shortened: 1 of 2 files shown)


granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -70,4 +72,29 @@ def find_v2_hierarchy() -> Optional[str]:
...
+        return v1_paths[controller]
+    if v2_path:
+        # v2 (unified) - all controllers are in a single unified hierarchy
+        return v2_path
+    # No cgroup mount of the requested controller
+    return None
+
+
+def is_valid_controller(controller: str) -> bool:
+    return controller in SUBSYSTEMS
```
d3dave, line 100: "valid" is a bit misleading. Perhaps "known"? This also does not check if the controller is actually in use on the system, so it's unclear what constitutes an "invalid" controller. A list of enabled subsystems is in `/proc/cgroups` if we actually care about that.

IlayRosenberg: Sure I can go for `known`. This isn't supposed to check if the controller is in use (mounted), but rather if the given string is a recognized cgroup controller

Jongy:
@d3dave - will you take the full CR here?

d3dave:
> @d3dave - will you take the full CR here?

Yes,  it already looks good. Just the func name and it's approved

Jongy:
> > @d3dave - will you take the full CR here?
> 
> Yes, it already looks good. Just the func name and it's approved

Thanks

d3dave (approved):

PR #104, opened 2022-11-29
glogger: Support replacing the glogger uri after initialization

netaneld122:


glogger/sender.py
```diff
@@ -58,19 +58,22 @@ def __init__(
...
-        self.server_uri = f"{scheme}://{server_address}/api/v1/logs"
+        self.set_server_uri(scheme, server_address)
         self.jsonify = JSONEncoder(separators=(",", ":"), default=repr).encode  # compact, no whitespace
         self.session = Session()
         self.messages_buffer: Optional[MessagesBuffer] = None
         self.metadata_callback: Callable[[], Dict] = lambda: {}
 
+    def set_server_uri(self, scheme: str, server_address: str) -> None:
+        self.server_address = server_address
+        self.server_uri = f"{scheme}://{server_address}/api/v1/logs"
```
d3dave, line 75: Instead of mutating `server_uri`, why not replace the sender on the handler with a new instance of Sender? Reason is to prevent the same send thread from sending to different servers in case `_send_loop` implements logic that only applies when sending to the same server. For example, we previously had implemented retries for failed requests, and retries don't make sense if the server address changes while the send function is executing.

Alternatively, I'd expect that we stop the thread before changing the `server_uri`, and then start it again. Or disallow calling `set_server_uri` while the thread is started (so the caller is responsible for the restart).

netaneld122: Yeah that's probably a better idea.

Looks like there's some additional state - such as `self.messages_buffer` that needs to be handled in this case.

netaneld122:
I'm closing this as I'm not going to touch glogger itself eventually

PR #105, opened 2022-11-30
Run cloud metadata queries in host net NS

Jongy:
I still need to see precisely why, but it seems that on some AWS instances (and possibly some Docker settings), a running container such as gProfiler is inaccessible from its net NS to the cloud metadata endpoint. I have some AWS instance now where gProfiler started not in host net NS fails to get cloud info, starting it with `--net=host` succeeds in getting cloud info, and I tested this branch and it succeeds as well.

I took some care to avoid wrapping unnecessary parts in `run_in_ns` (e.g the `logger` call which for all we know can trigger networking operations). The smallest section possible would be in `send_request`, but I think what I did is good enough

### DIFF
--- a/granulate_utils/metadata/cloud.py
+++ b/granulate_utils/metadata/cloud.py
@@ -6,13 +6,14 @@
 import logging
 from dataclasses import dataclass
 from http.client import NOT_FOUND
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import requests
 from requests import Response
 from requests.exceptions import ConnectionError
 
 from granulate_utils.exceptions import BadResponseCode
+from granulate_utils.linux.ns import run_in_ns
 from granulate_utils.metadata import Metadata
 
 METADATA_REQUEST_TIMEOUT = 5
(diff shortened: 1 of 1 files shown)


Jongy:
@DanielShaulov - I ran old gProfiler 1.5.0 on the same server, and it succeeded in getting the metadata. Which leads to the understanding that API v1 works inside containers but v2 doesn't, at least in some cases.

DanielShaulov:
@Jongy - can you run the `curl` that you ran to test it outside, inside the container to cement this fact? It seems very odd that different APIs would have different routing inside/outside of container.

DanielShaulov (approved):

Jongy:
The request to get the token blocks in a container and succeeds on the host: `curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 120"`

Jongy:
```
*   Trying 169.254.169.254:80...
* Connected to 169.254.169.254 (169.254.169.254) port 80 (#0)
> PUT /latest/api/token HTTP/1.1
> Host: 169.254.169.254
> User-Agent: curl/7.81.0
> Accept: */*
> X-aws-ec2-metadata-token-ttl-seconds: 120
> 
```

PR #106, opened 2022-12-01
granulate_utils: Handle unlimited cgroup quota

mirikl:
Take into account that a cgroup can be unlimited (`cfs_quota_us` is set to -1).

### DIFF
--- a/granulate_utils/linux/cgroups/cpu_cgroup.py
+++ b/granulate_utils/linux/cgroups/cpu_cgroup.py
@@ -19,7 +19,9 @@ class CpuCgroup(BaseCgroup):
 
     def get_cpu_limit_cores(self) -> float:
         period = int(self.read_from_control_file(self.cfs_period_us))
-        return int(self.read_from_control_file(self.cfs_quota_us)) / period
+        quota = int(self.read_from_control_file(self.cfs_quota_us))
+        # if quota is set to -1 it means this cgroup is unlimited
+        return quota / period if quota != -1 else -1.0
 
     def reset_cpu_limit(self) -> None:
         self.write_to_control_file(self.cfs_quota_us, "-1")
--- a/tests/granulate_utils/test_cgroups.py
+++ b/tests/granulate_utils/test_cgroups.py
@@ -4,15 +4,15 @@ from unittest.mock import patch
 from granulate_utils.linux.cgroups.cpu_cgroup import CpuCgroup
 
 
-def wrap_read_from_control_file(mapping: Dict[str, str]):
-    def func(tunable_name: str):
+def wrap_read_from_control_file(mapping: Dict[str, str]) -> Callable:
+    def func(tunable_name: str) -> str:
         return mapping[tunable_name]
 
     return func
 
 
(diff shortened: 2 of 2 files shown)


tests/granulate_utils/test_cgroups.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def wrap_read_from_control_file(mapping: Dict[str, str]) -> Callable:
+    def func(tunable_name: str) -> str:
+        return mapping[tunable_name]
+
+    return func
+
+
+@patch("granulate_utils.linux.cgroups.cpu_cgroup.BaseCgroup.__init__", return_value=None)
```
Jongy, line 14: Maybe patch only `_verify_preconditions`? (I figure that's the reason behind this patch)

Jongy (changes requested):
1 small, and in any case I'll let @michelhe decide if that's the API we want here.

granulate_utils/linux/cgroups/cpu_cgroup.py
```diff
@@ -19,7 +19,9 @@ def set_cpu_limit_cores(self, cores: float) -> None:
 
     def get_cpu_limit_cores(self) -> float:
         period = int(self.read_from_control_file(self.cfs_period_us))
-        return int(self.read_from_control_file(self.cfs_quota_us)) / period
+        quota = int(self.read_from_control_file(self.cfs_quota_us))
+        # if quota is set to -1 it means this cgroup is unlimited
+        return quota / period if quota != -1 else -1.0
```
michelhe, line 24: It would be nicer to return `None` instead of `-1` but linux likes using `-1` for these tunables
I'm fine with keeping this as `-1` to represent unlimited

d3dave: that makes sense when you are writing in C and using the sign bit to represent a null value, but we are in python here so testing for `< 0` is more confusing than testing `is None`.

tests/granulate_utils/test_cgroups.py
```diff
@@ -0,0 +1,39 @@
...
+from unittest.mock import patch
+
+from granulate_utils.linux.cgroups.cpu_cgroup import CpuCgroup
+
+
+def wrap_read_from_control_file(mapping: Dict[str, str]) -> Callable:
+    def func(tunable_name: str) -> str:
+        return mapping[tunable_name]
+
+    return func
```
d3dave, line 11: can just do `return mapping.__getitem__`

granulate_utils/linux/cgroups/cpu_cgroup.py
mirikl: @d3dave , I'd rather not change the return value to `None` as it changes the API since the function will return `Optional[Float]` instead of `float`. I don't mind changing it but I'd like to talk offline about what else I need to check for this change as I'm not entirely sure.

d3dave: Ok. No real issue with keeping it -1.0. The only two callsites I found don't test the value anyway.

d3dave (approved):

PR #107, opened 2022-12-01
Drop Python 3.6, 3.7 support

Jongy:
With https://github.com/Granulate/gprofiler/pull/597 merged, we no longer use 3.6 (and nothing uses 3.7). Now the minimum version is 3.8.

I will give some time to the version based on https://github.com/Granulate/gprofiler/pull/597 to run (to make sure we don't go back to 3.6) before I change it.

### DIFF
--- a/.github/workflows/glogger.yml
+++ b/.github/workflows/glogger.yml
@@ -6,7 +6,7 @@ jobs:
   lint:
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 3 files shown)


Jongy:
@d3dave let's merge this first, seems like `actions/setup-python` fails for 3.6 on other PRs now

Jongy:
I am good with dropping support of 3.6, 3.7

michelhe (approved):

PR #109, opened 2022-12-07
linux: Use /proc/tid/mountinfo instead of /proc/self/mountinfo

d3dave:
When running in a thread which is in a different namespace, `/proc/self/...` will return the information from the process's namespaces instead of the thread's namespaces.

### DIFF
--- a/granulate_utils/linux/mountinfo.py
+++ b/granulate_utils/linux/mountinfo.py
@@ -1,9 +1,8 @@
 """
 See section 3.5 in https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 """
-from typing import Iterable, List, NamedTuple, Union
-
-from typing_extensions import Literal
+import threading
+from typing import Iterable, List, NamedTuple, Optional
 
 
 class Mount(NamedTuple):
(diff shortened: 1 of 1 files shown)


Jongy (approved):
I don't recall if all files exist under `thread-self`, assuming u have checked this one

Jongy:
See https://github.com/Granulate/granulate-utils/pull/107 about CI failure

michelhe (dismissed):

Jongy (approved):

michelhe:
> @Jongy @michelhe says we want to support Linux <3.17 (CentOS 7 or some such), where `thread-self` is unavailable. wdys?

Instead of `thread-self` we can just do `/proc/{pid}/task/{tid}`

michelhe:
> > > > @Jongy @michelhe says we want to support Linux <3.17 (CentOS 7 or some such), where `thread-self` is unavailable. wdys?
> > > 
> > > 
> > > Instead of `thread-self` we can just do `/proc/{pid}/task/{tid}`
> > 
> > 
> > Let's use that
> 
> For this we need gettid(), which is not exposed.

And introduce another use of `ctypes` is not an option, right ? :(

Jongy:
https://docs.python.org/3/library/threading.html#threading.get_native_id

d3dave:
Verified this works on Amzn Linux 2018.

Jongy (approved):
Can fix commit title to `Use "/proc/tid/" instead of "/proc/thread-self"`

PR #112, opened 2022-12-21
java: JVM version parsing tests

Jongy:


uriariel (approved):

.github/workflows/glogger.yml
```diff
@@ -3,34 +3,6 @@ name: gLogger
 on: pull_request
 
 jobs:
-  lint:
```
Jongy, line 6: This is exactly like the granulate-utils one, so I removed it here.

PR #113, opened 2022-12-22
java: Consolidate VM type & Zing major version parsing code

Jongy:


granulate_utils/java.py
```diff
@@ -114,11 +114,16 @@ def java_exit_code_to_signo(exit_code: int) -> Optional[int]:
...
+VmType = Literal["HotSpot", "Zing", "OpenJ9", None]
+
+
 @dataclass
 class JvmVersion:
     version: Version
     build: int
     name: str
+    vm_type: VmType
+    zing_major: Optional[int] = None  # non-None if Zing
```
YoniKF, line 126: I'm not sure that it's supported, but consider adding
``` python
@dataclass
class ZingJvmVersion(JvmVersion):
    zing_major: int
```

YoniKF (approved):
One stylistic suggestion, you can resolve and merge if you are not interested in it.

granulate_utils/java.py
Jongy: It is but it won't help much, vm type checking I still want to do generic with `.vm_type == ...` so then I'll have to cast objects to access the field with linter permission

PR #114, opened 2022-12-22
linux: process: Handle errors in _read_process_auxv()

Jongy:
Fixes this case:
```
File "granulate_utils.linux.process", line 77, in read_process_execfn
File "granulate_utils.linux.process", line 93, in _read_process_auxv
ValueError: auxv id 31 was not found!
```

which happens when the function is called on kernel threads / zombie processes.

### DIFF
--- a/granulate_utils/linux/process.py
+++ b/granulate_utils/linux/process.py
@@ -81,6 +81,13 @@ def read_process_execfn(process: psutil.Process) -> str:
 
 def _read_process_auxv(process: psutil.Process, auxv_id: int) -> int:
     auxv = read_proc_file(process, "auxv")
+    if not auxv:
+        # kernel thread / zombie process yield this result.
+        # we don't expect to be called on kernel threads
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/process.py
```diff
@@ -81,6 +81,13 @@ def read_process_execfn(process: psutil.Process) -> str:
 
 def _read_process_auxv(process: psutil.Process, auxv_id: int) -> int:
     auxv = read_proc_file(process, "auxv")
+    if not auxv:
+        # kernel thread / zombie process yield this result.
```
netaneld122, line 85: Can't it be dead as well?

d3dave: If it's dead, there's no proc file.

Jongy: Yup, we'll get `NoSuchProcess` which is also good.

Callers of this function will need to handle `NoSuchProcess` & `ZombieProcess` the way the please.

PR #115, opened 2022-12-27
normpath-resolve-proc-root-links

guybortnikov:
returnes path may contain relative symlinks which might contain `.` or `..`, `normpath` "eliminates" them by "resolving" these simple built in links


Jongy (changes requested):
Still thinking

PR #117, opened 2023-01-02
Refactor CGroup utils

mirikl:
Change objects to allow setting controller attributes for Cgroups other than the current process Cgroup.


mirikl:
Need to add tests

michelhe:
If you are already moving modules around, I much like `from granulate_utils.linux.cgroups.controllers import CpuController, MemoryController` vs  ``from granulate_utils.linux.cgroups.cpu_controller import CpuController ; ...`

mirikl:
> If you are already moving modules around, I much like `from granulate_utils.linux.cgroups.controllers import CpuController, MemoryController` vs ``from granulate_utils.linux.cgroups.cpu_controller import CpuController ; ...`

I'd still rather have them in separate files if you don't mind, but I can add a controllers file which will have all of the imports in it. WDYT?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -98,3 +99,67 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
+    :return: the relative path, with a '/' prefix
+    """
+    hierarchy_details = get_cgroups()
+    for line in hierarchy_details:
+        if subsystem in line[1]:
+            return line[2]
+    raise Exception(f"{subsystem!r} not found")
+
+
+class CgroupUtils:
```
d3dave, line 117: Why not define all the methods top-level?

Jongy: :+1: , no value for classes if it's functional after all.

Jongy:
I'll review this tmrw

granulate_utils/linux/cgroups/base_controller.py
```diff
@@ -0,0 +1,54 @@
...
+    get_current_process_cgroup,
+    is_known_controller,
+)
+
+
+class BaseController:
+    controller: str
+
+    def __init__(self, cgroup: Optional[Union[Path, CgroupCore]] = None) -> None:
+        assert is_known_controller(self.controller), f"{self.controller!r} is not supported"
```
Jongy, line 23: `self.controller` was no initialized at this point?

Jongy: Oh, this is a class attribute?

Jongy: Maybe comment for clarity

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +197,39 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
+            return cgroup_v2
     # No cgroup mount of the requested controller
     return None
 
 
-def is_known_controller(controller: str) -> bool:
-    return controller in SUBSYSTEMS
+def create_cgroup_from_path(controller: str, cgroup_path: Path) -> CgroupCore:
+    cgroup_mount = get_cgroup_mount(controller)
+    assert cgroup_mount is not None, "Could not find cgroup mount point"
```
Jongy, line 212: This assert repeats multiple times, move to `get_cgroup_mount` or if you want a variant that can fail, create a `get_cgroup_mount_checked`

Also, add the `controller` to the assert message

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +197,39 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
+
+    try:
```
Jongy, line 214: Explain what's done here?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -28,32 +34,43 @@
 }
 
 
-def get_cgroups(process: Optional[psutil.Process] = None) -> List[Tuple[str, List[str], str]]:
+@dataclass
+class ProcCgroup:
+    hier_id: str
+    cgroup_relative_path: str
+    controllers: List[str]
```
d3dave, line 41: `cgroup_relative_path` is incorrect terminology. The path _is_ the cgroup. Also, the order should be id, controllers, path. Refer to `cgroup(7)`:

>The colon-separated fields are, from left to right:
>
>1. For cgroups version 1 hierarchies, this field contains a unique hierarchy ID number that can be matched to a hierarchy ID in /proc/cgroups.  For the cgroups version 2 hierarchy, this field contains the value 0.
>
>2. For cgroups version 1 hierarchies, this field contains a comma-separated list of the controllers bound to the hierarchy.  For the cgroups version 2 hierarchy, this field is empty.
>
>3. This field contains the pathname of the control group in the hierarchy to which the process belongs.  This pathname is relative to the mount point of the hierarchy.

d3dave (changes requested):

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -28,32 +34,43 @@
 }
 
 
-def get_cgroups(process: Optional[psutil.Process] = None) -> List[Tuple[str, List[str], str]]:
+@dataclass
+class ProcCgroup:
```
Jongy, line 38: `ProcCgroupLine`? And drop a comment showing an example line?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +95,99 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+    path: Path
+
+    def __init__(self, path: Path):
+        self.path = path
+
+    def _get_parent_cgroup(self, cgroup_name: str) -> Optional[Path]:
+        cgroup_index = self.path.parts.index(cgroup_name)
+        if cgroup_index != -1:
+            return Path(*self.path.parts[: cgroup_index + 1])
+        return None
```
Jongy, line 111: Can be written more concisely with `split`

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +95,99 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+
+    def _create_subcgroup(self, cgroup_name: str) -> Optional[Path]:
+        if self.has_parent_cgroup(cgroup_name):
+            new_cgroup_path = self._get_parent_cgroup(cgroup_name)
+        else:
+            new_cgroup_path = self.path / cgroup_name
+            new_cgroup_path.mkdir(exist_ok=True)
```
Jongy, line 118: Something in the terminology here doesn't make sense. A function called "create_subgroup" checks if the group has a parent and it so returns the parent, otherwise create a child. ???

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +95,99 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+
+        return new_cgroup_path
+
+    def has_parent_cgroup(self, cgroup_name: str) -> bool:
```
Jongy, line 122: I'd avoid re-implementing the "parts" logic and implement this function as `return _get_parent_cgroup(name) is not None`

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +95,99 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+        """
+        Check if cgroup has the given cgroup_name in its hierarchy
+        :param cgroup_name: the cgroup name
+        :return: bool indicating if cgroup is in hierarchy
+        """
+        controller_path_parts = self.path.parts
+        return cgroup_name in controller_path_parts
+
+    def get_pids_in_cgroup(self) -> set[int]:
+        return {int(proc) for proc in self.read_from_interface_file(CGROUP_PROCS_FILE).split()}
```
Jongy, line 132: I know this file is V2, is it also supported in V1?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +95,99 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+
+    def assign_process_to_cgroup(self, pid: int = 0) -> None:
+        if not self.path.exists():
+            raise FileNotFoundError("Cgroup doesn't exist")
```
Jongy, line 136: 1. Create a specific exception type for it. Personally I dislike raising builtin exceptions for non-builtin cases, especially when the message is modified and the meaning is different
2. Don't check and raise then run the opreation. Just run the operation and let the exception pass up, OR catch and re-raise appropriately. Checking-raising-writing is racy (what if file goes down between the check and the write?) and error prone in general (what if the error can be raised in cases you are not checking before?)

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +95,99 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+
+class CgroupCoreV1(CgroupCore):
+    def create_subcgroup(self, controller: str, cgroup_name: str) -> CgroupCore:
+        subcgroup_path = self._create_subcgroup(cgroup_name)
+        assert subcgroup_path is not None
+        return CgroupCoreV1(subcgroup_path)
+
+
+CGROUP_SUPPORTED_CONTROLLERS = "cgroup.controllers"
+CGROUP_ENABLED_CONTROLLERS = "cgroup.subtree_control"
```
Jongy, line 165: Stuff that are only V2 - let's make sure V2 is included in their name, e.g prefix with `CGROUP_V2_...`

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +95,99 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+
+
+class CgroupCoreV2(CgroupCore):
+    def is_controller_supported(self, controller: str):
+        return controller in self.read_from_interface_file(CGROUP_SUPPORTED_CONTROLLERS).split()
+
+    def enable_controller(self, controller: str):
+        assert self.is_controller_supported(controller), "Controller not supported"
```
Jongy, line 173:
```suggestion
        assert self.is_controller_supported(controller), f"Controller not supported {controller!r}"
```
always give the bad data name in exception. this ain't pytest, you won't see the name otherwise

granulate_utils/linux/cgroups/cgroup.py
d3dave: docs say that it is, apparently

mirikl: They also removed the `tasks` file

granulate_utils/linux/cgroups/cgroup.py
mirikl: I don't want to have 2 different functions to create a CgroupCore object from a path (one for relative path from the mount point and another for full path). So I'm checking if the given path is relative to the mount point and if not I prefix the path with it. The problem is that we don't have python 3.9 `is_relative_to` so I'm doing it with try catch like the pathlib impl does.

granulate_utils/linux/cgroups/base_controller.py
mirikl: yes it's a class attribute, i'll add a comment

granulate_utils/linux/cgroups/cgroup.py
mirikl: yeah you're right it is racey, I'm removing it

granulate_utils/linux/cgroups/cgroup.py
mirikl: yeah subcgroup doesn't really explain it, what about `get_cgroup_in_hierarchy` ?

granulate_utils/linux/cgroups/cgroup.py
Jongy: Understood

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +198,46 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
-def is_known_controller(controller: str) -> bool:
-    return controller in SUBSYSTEMS
+def get_cgroup_mount_checked(controller: str, resolve_host_root_links: bool = True) -> CgroupCore:
+    cgroup_mount = get_cgroup_mount(controller)
+    assert cgroup_mount is not None, f"Could not find cgroup mount point for controller {controller!r}"
+    return cgroup_mount
+
+
+def create_cgroup_from_path(controller: str, cgroup_path: Path) -> CgroupCore:
+    cgroup_mount = get_cgroup_mount_checked(controller)
```
Jongy, line 218: Glossary wise - let's create consistent naming here. `cgroup_path` here might be relative to the controller and might already include the controller & mount, right?

Let's create standard names:
* `cgroup_full_path` - mount+controller+path
* `cgroup_path` - path on top of the controller mount
* `controller` - just the controller name
* `cgroup_mount` - just the mount
Makes sense?

Then this function becomes:
```
def create_cgroup_from_path(controller: str, cgroup_path_or_full_path: Path):
    try:
        cgroup_path_or_full_path.relative_to(cgroup_mount.path)
        # it's a full path
        cgroup_full_path = cgroup_path_or_full_path
    except ValueError:
        # it's path relative to the controller
        cgroup_full_path = cgroup_mount.path / cgroup_path
```

And fix other functions accordingly. Makes it easier to browse through the code if you use consistent names, since things don't have types and a `Path` can be either.

Alternatively, you can create types for them... but I think names suffice.

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +198,46 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
+
+    try:
+        cgroup_path.relative_to(cgroup_mount.path)
+    except ValueError:
+        cgroup_path = cgroup_mount.path / cgroup_path
+
+    if isinstance(cgroup_mount, CgroupCoreV1):
```
Jongy, line 225: API-wise - I think `cgroup_mount.is_v1` property is nicer

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +198,46 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
+    """
+    Get a CgrouopCore object for a given process. If process is None return for current process.
+    """
+    cgroup_mount = get_cgroup_mount_checked(controller)
+    is_v1 = isinstance(cgroup_mount, CgroupCoreV1)
+
+    for process_cgroup in get_process_cgroups(process):
+        if is_v1 and controller in process_cgroup.controllers:
+            return CgroupCoreV1(cgroup_mount.path / process_cgroup.relative_path[1:])
+        elif not is_v1 and process_cgroup.hier_id == "0":
```
Jongy, line 241: Can `hier_id` ever be not 0 for v2?

If not - remove the `if` and instead `assert` it.

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +198,46 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
+def get_cgroup_for_process(controller: str, process: Optional[psutil.Process] = None) -> CgroupCore:
+    """
+    Get a CgrouopCore object for a given process. If process is None return for current process.
+    """
+    cgroup_mount = get_cgroup_mount_checked(controller)
+    is_v1 = isinstance(cgroup_mount, CgroupCoreV1)
+
+    for process_cgroup in get_process_cgroups(process):
+        if is_v1 and controller in process_cgroup.controllers:
+            return CgroupCoreV1(cgroup_mount.path / process_cgroup.relative_path[1:])
```
Jongy, line 240: `[1:]` is to remove the leading slash?

Instead `lstrip("/")`?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+    def read_from_interface_file(self, interface_name: str) -> str:
+        interface_path = self.path / interface_name
+        return interface_path.read_text()
+
+    def write_to_interface_file(self, interface_name: str, data: str) -> None:
+        interface_path = self.path / interface_name
+        interface_path.write_text(data)
+
+    @abstractmethod
+    def get_cgroup_in_hierarchy(self, controller: str, cgroup_name: str) -> CgroupCore:
```
Jongy, line 150: Name does not signify the difference from `has_parent_cgroup`.

This is for our specific usage of creating a cgroup under the existing hierarchy OR use the one that our process already resides in, correct? I'd use a more specific name then: `get_or_create_cgroup`, something in that area

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+        Check if cgroup has the given cgroup_name in its hierarchy
+        :param cgroup_name: the cgroup name
+        :return: bool indicating if cgroup is in hierarchy
+        """
+        return self._get_parent_cgroup(cgroup_name) is not None
+
+    def get_pids_in_cgroup(self) -> set[int]:
+        return {int(proc) for proc in self.read_from_interface_file(CGROUP_PROCS_FILE).split()}
+
+    def assign_process_to_cgroup(self, pid: int = 0) -> None:
```
Jongy, line 138: In docstring - PID 0 means current process/thread

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
     return path
 
 
-def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> Optional[str]:
+CGROUP_PROCS_FILE = "cgroup.procs"
+
+
+class CgroupCore:
+    path: Path
```
Jongy, line 108: Docstring comment for it - the `cgroup_path` per the naming I suggested

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+CGROUP_V2_SUPPORTED_CONTROLLERS = "cgroup.controllers"
+CGROUP_V2_ENABLED_CONTROLLERS = "cgroup.subtree_control"
+
+
+class CgroupCoreV2(CgroupCore):
+    def is_controller_supported(self, controller: str):
+        return controller in self.read_from_interface_file(CGROUP_V2_SUPPORTED_CONTROLLERS).split()
+
+    def enable_controller(self, controller: str):
+        assert self.is_controller_supported(controller), f"Controller not supported {controller!r}"
```
Jongy, line 174: To my understand, here you're protecting from this case:
image

Please, instead of trying to mimic the protection logic that cgroup employs in the kernel, you just try to do things & catch the related errors and handle them per their meaning. In this case, you're `assert`ing which means you don't expect this to happen - so instead you can `try` and if you get `ENOENT` you narrow it down to `ENOENT` from the `write` call itself (perhaps separate to `open` and `write`) and THEN you raise an exception which tells the controller is disabled.

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+        enabled_controllers_file = self.path / CGROUP_V2_ENABLED_CONTROLLERS
+        enabled_controllers_file.write_text(f"+{controller}{os.linesep}")
```
Jongy, line 176: Is `os.linesep` really needed? I'd say it's not - since you write and close the fd, sysfs understands it as "that's all the input"

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+
+    def disable_controller(self, controller: str):
```
Jongy, line 178: Btw all `controller: str` can be replaced to `controller: ControllerType` where `ControllerType = Literal["....", ..., ...]`. Just make sure it's supported in python 3.8 as well; basically if CI passes, you're good.

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+        enabled_controllers_file = self.path / CGROUP_V2_ENABLED_CONTROLLERS
+        enabled_controllers_file.write_text(f"-{controller}{os.linesep}")
+
+    def get_cgroup_in_hierarchy(self, controller: str, cgroup_name: str) -> CgroupCore:
+        subcgroup_path = self._get_cgroup_in_hierarchy(cgroup_name)
+        assert subcgroup_path is not None
+
+        new_cgroup = CgroupCoreV2(subcgroup_path)
+        new_cgroup.enable_controller(controller)
+        return new_cgroup
```
Jongy, line 188: So this, logically, will either enable a controller for our current `cgroup_path`, or will enable it for a `cgroup_path` that was created under us.

Is this comment relevant?
image

i.e, if the CURRENT cgroup don't have the controller enabled, then a child cgroup won't be able to enable it, to my understanding.

granulate_utils/linux/cgroups/cgroup.py
Jongy: I gave a suggestion in a different comment, letting the discussion move there.

granulate_utils/linux/cgroups/cgroup.py
Jongy: Same as I said here https://github.com/Granulate/granulate-utils/pull/117#discussion_r1066443110

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -3,15 +3,20 @@
...
+from typing import List, Mapping, Optional
 
 import psutil
 
 from granulate_utils.linux import ns
 from granulate_utils.linux.mountinfo import iter_mountinfo
 from granulate_utils.linux.process import read_proc_file
 
-SUBSYSTEMS = {
+CONTROLLERS = {
```
Jongy, line 19: The list is different between V2 and V1, how does it work here with you having a single list?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -28,32 +33,50 @@
...
 
-def get_cgroups(process: Optional[psutil.Process] = None) -> List[Tuple[str, List[str], str]]:
+class ProcCgroupLine:
     """
-    Get the cgroups of a process in [(hier id., controllers, path)] parsed form.
-    If process is None, gets the cgroups of the current process.
+    The format of the line:  hierarchy-ID:controller-list:cgroup-path
+    Example line: 1:cpu:/custom_cgroup
+
+    cgroup-path - the cgroup the process belongs to, relative to the hierarchy mount point e.g. /sys/fs/cgroup
```
Jongy, line 41:
```suggestion
    cgroup-path - the cgroup the process belongs to, relative to the hierarchy mount point e.g. /sys/fs/cgroup/memory on v1 or just the cgroups v2 mount on v2 e.g /sys/fs/cgroup.
```

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -28,32 +33,50 @@
...
+    Get the cgroups of a process in [(hier id., controllers, path)] parsed form.
+    If process is None, gets the cgroups of the current process.
+    """
     process = process or psutil.Process()
     text = read_proc_file(process, "cgroup").decode()
-    return [parse_line(line) for line in text.splitlines()]
+    return [ProcCgroupLine(line) for line in text.splitlines()]
+
+
+def is_known_controller(controller: str) -> bool:
```
Jongy, line 65: Separate for v1 v2

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -28,32 +33,50 @@
...
 
 def find_v1_hierarchies(resolve_host_root_links: bool = True) -> Mapping[str, str]:
     """
     Finds all the mounted hierarchies for all currently enabled cgroup v1 controllers.
-    :return: A mapping from cgroup subsystem names to their respective hierarchies.
+    :return: A mapping from cgroup controller names to their respective hierarchies.
     """
     hierarchies = {}
     for mount in iter_mountinfo(1):
         # The mount source is always "cgroup" by convention, we don't really have to check it.
```
Jongy, line 76: I don't understand the comment. I mean, I do, but why put it here. Let's remove?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -28,32 +33,50 @@
...
         if mount.mount_source != "cgroup" or mount.filesystem_type != "cgroup":
             continue
-        controllers = set(mount.super_options) & SUBSYSTEMS
+        controllers = set(mount.super_options) & CONTROLLERS
```
Jongy, line 79: Can also `assert set(mount.super_options).intersect(CONTROLLERS) == set(mount.super_options)` as a self-check, otherwise we're missing controllers..

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -28,32 +33,50 @@
...
     text = read_proc_file(process, "cgroup").decode()
-    return [parse_line(line) for line in text.splitlines()]
+    return [ProcCgroupLine(line) for line in text.splitlines()]
+
+
+def is_known_controller(controller: str) -> bool:
+    return controller in CONTROLLERS
 
 
 def find_v1_hierarchies(resolve_host_root_links: bool = True) -> Mapping[str, str]:
```
Jongy, line 69: As part of your PR let's do an API cleanup. Why does `resolve_host_root_links` exist? I don't see it ever set in our codebase. This is always true, you never want it "false" afaict

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
     return path
 
 
-def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> Optional[str]:
+CGROUP_PROCS_FILE = "cgroup.procs"
```
Jongy, line 104: Comment that it exists on both v1 v2

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+        self.path = path
+
+    def _get_parent_cgroup(self, cgroup_name: str) -> Optional[Path]:
+        if cgroup_name in self.path.parts:
+            return Path(self.path.as_posix().split(cgroup_name)[0]) / cgroup_name
+        return None
+
+    def _get_cgroup_in_hierarchy(self, cgroup_name: str) -> Optional[Path]:
+        if self.has_parent_cgroup(cgroup_name):
+            new_cgroup_path = self._get_parent_cgroup(cgroup_name)
```
Jongy, line 120: Btw, is there any expected case where this will return an ancestor cgroup? In normal cases (unless we have bugs) it'll always be CURRENT cgroup or no cgroup exists and then we create a child, right?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+    def get_cgroup_in_hierarchy(self, controller: str, cgroup_name: str) -> CgroupCore:
+        """
+        Return a CGroup in the hierarchy which has the given name.
+        If one doesn't exist higher up in the hierarchy, create a new one lower in the hierarchy.
+        """
+        pass
+
+
+class CgroupCoreV1(CgroupCore):
+    def get_cgroup_in_hierarchy(self, controller: str, cgroup_name: str) -> CgroupCore:
```
Jongy, line 159: `assert controller in CONTROLLERS, controller` just for a clear API.

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+
+    def __init__(self, path: Path):
+        self.path = path
+
+    def _get_parent_cgroup(self, cgroup_name: str) -> Optional[Path]:
+        if cgroup_name in self.path.parts:
+            return Path(self.path.as_posix().split(cgroup_name)[0]) / cgroup_name
+        return None
+
+    def _get_cgroup_in_hierarchy(self, cgroup_name: str) -> Optional[Path]:
```
Jongy, line 118: This function ain't really `Optional`, mypy cannot see through `_get_parent_cgroup` returning non-None the 2nd time you call it, after `has_parent_cgroup` returned true. Instead of using an incorrect type, you should help mypy by adding an `assert new_cgroup_path is not None` after the 2nd `_get_parent_cgroup` call.

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+        Return a CGroup in the hierarchy which has the given name.
+        If one doesn't exist higher up in the hierarchy, create a new one lower in the hierarchy.
+        """
+        pass
+
+
+class CgroupCoreV1(CgroupCore):
+    def get_cgroup_in_hierarchy(self, controller: str, cgroup_name: str) -> CgroupCore:
+        subcgroup_path = self._get_cgroup_in_hierarchy(cgroup_name)
+        assert subcgroup_path is not None
```
Jongy, line 161: This assert doesn't fit here, remove after you do https://github.com/Granulate/granulate-utils/pull/117/files#r1089819098

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+        enabled_controllers_file = self.path / CGROUP_V2_ENABLED_CONTROLLERS
+        enabled_controllers_file.write_text(f"+{controller}{os.linesep}")
+
+    def disable_controller(self, controller: str):
+        enabled_controllers_file = self.path / CGROUP_V2_ENABLED_CONTROLLERS
+        enabled_controllers_file.write_text(f"-{controller}{os.linesep}")
+
+    def get_cgroup_in_hierarchy(self, controller: str, cgroup_name: str) -> CgroupCore:
+        subcgroup_path = self._get_cgroup_in_hierarchy(cgroup_name)
+        assert subcgroup_path is not None
```
Jongy, line 184: Remove assert https://github.com/Granulate/granulate-utils/pull/117/files#r1089819098

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+
+        new_cgroup = CgroupCoreV2(subcgroup_path)
+        new_cgroup.enable_controller(controller)
+        return new_cgroup
+
+
+def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> Optional[CgroupCore]:
```
Jongy, line 191: Pls check here as well, if `resolve_host_root_links` is truly needed.

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +198,46 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
     v2_path = find_v2_hierarchy(resolve_host_root_links)
     if v1_paths and controller in v1_paths:
```
Jongy, line 199:
```suggestion
    if controller in v1_paths:
```

Jongy: Btw is this check correct? For a controller I suppose we allow `cpu` and the underlying v1 controller might be `cpu,cpuacct`; what will happen in this case?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +198,46 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
         # Either v1 or hybrid with the requested controller bound to a v1 hierarchy
-        return v1_paths[controller]
+        return CgroupCoreV1(Path(v1_paths[controller]))
     if v2_path:
-        # v2 (unified) - all controllers are in a single unified hierarchy
-        return v2_path
+        # v2 (unified) - check given controller is supported in cgroup v2
+        cgroup_v2 = CgroupCoreV2(Path(v2_path))
+        if cgroup_v2.is_controller_supported(controller):
```
Jongy, line 205: This will happen if the particular controller is already controlled by the parent cgroup, right? In that case I'd say we want a specific exception and NOT a behavior of "controller is not available", which is implied by this function returning None.

Jongy: Also, the comment below `# No cgroup mount of the requested controller` is misleading because that's not the case here.

Jongy: I see it's checked in `get_cgroup_mount_checked`, commenting there as well

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +198,46 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
+            return cgroup_v2
     # No cgroup mount of the requested controller
     return None
 
 
-def is_known_controller(controller: str) -> bool:
-    return controller in SUBSYSTEMS
+def get_cgroup_mount_checked(controller: str, resolve_host_root_links: bool = True) -> CgroupCore:
+    cgroup_mount = get_cgroup_mount(controller)
+    assert cgroup_mount is not None, f"Could not find cgroup mount point for controller {controller!r}"
```
Jongy, line 213: This checks what I commented [here](https://github.com/Granulate/granulate-utils/pull/117/files#r1089819591) I suppose. I'd say it's not an `assert` - asserts are for things which are bugs in your code, while this case ain't a bug - if our process runs in a cgroup where the parent cgroup already controls the controller we wish... we should be notified of it & fix the problem, possibly automatically in the code.

OR I misunderstood this concept, will comment here back if I understand it better, or you let me know if I'm wrong

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -88,13 +198,46 @@ def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> O
...
+    Get a CgrouopCore object for a given process. If process is None return for current process.
+    """
+    cgroup_mount = get_cgroup_mount_checked(controller)
+    is_v1 = isinstance(cgroup_mount, CgroupCoreV1)
+
+    for process_cgroup in get_process_cgroups(process):
+        if is_v1 and controller in process_cgroup.controllers:
+            return CgroupCoreV1(cgroup_mount.path / process_cgroup.relative_path[1:])
+        elif not is_v1 and process_cgroup.hier_id == "0":
+            return CgroupCoreV2(cgroup_mount.path / process_cgroup.relative_path[1:])
```
Jongy, line 242: `lstrip` as suggested

granulate_utils/linux/cgroups/cpu_controller.py
```diff
@@ -0,0 +1,110 @@
...
+    @abstractmethod
+    def get_cpu_limit_quota(self) -> int:
+        pass
+
+    @abstractmethod
+    def get_cpu_limit_params(self) -> CpuLimitParams:
+        pass
+
+    @abstractmethod
+    def set_cpu_limit_quota(self, quota: str) -> None:
```
Jongy, line 40: `quota: int` to be consistent with your getters

granulate_utils/linux/cgroups/cpu_controller.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self, cgroup: CgroupCore):
+        self.cgroup = cgroup
+
+    @abstractmethod
+    def get_cpu_limit_period(self) -> int:
+        pass
+
+    @abstractmethod
+    def get_cpu_limit_quota(self) -> int:
+        pass
```
Jongy, line 33: Document special values (`-1` only, I think?) in docstring

granulate_utils/linux/cgroups/cpu_controller.py
```diff
@@ -0,0 +1,110 @@
...
+    def get_cpu_limit_params(self) -> CpuLimitParams:
+        pass
+
+    @abstractmethod
+    def set_cpu_limit_quota(self, quota: str) -> None:
+        pass
+
+
+class CpuControllerV1(CpuControllerInterface):
+    cpu_stat = "cpu.stat"
```
Jongy, line 45:
```suggestion
    CPU_STAT_FILE = "cpu.stat"
```

for others as well. It's confusing that a `CpuController` object will have `self.quota` which is not its quota lol (even though it's a class attr, it's available via `self`, and it's const so make it CONST)

granulate_utils/linux/cgroups/cpu_controller.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@dataclass
+class CpuLimitParams:
+    period: int
+    quota: int
+
+
+class CpuControllerInterface:
+    cpu_stat: str
```
Jongy, line 22: `CPU_STAT_FILE`

granulate_utils/linux/cgroups/cpu_controller.py
```diff
@@ -0,0 +1,110 @@
...
+    def get_cpu_limit_period(self) -> int:
+        return self.get_cpu_limit_params().period
+
+    def get_cpu_limit_quota(self) -> int:
+        return self.get_cpu_limit_params().quota
+
+    def get_cpu_limit_params(self) -> CpuLimitParams:
+        cpu_limit = self.cgroup.read_from_interface_file(self.cpu_limit_file).split()
+        return CpuLimitParams(
+            period=int(cpu_limit[1]), quota=int(cpu_limit[0]) if cpu_limit[0] != self.UNBOUNDED_QUOTA_VALUE else -1
```
Jongy, line 76: Document that `-1` value is used to match v1

granulate_utils/linux/cgroups/cpu_controller.py
```diff
@@ -0,0 +1,110 @@
...
+        cpu_limit_params = self.controller_interface.get_cpu_limit_params()
+        # if quota is set to -1 it means this cgroup is unlimited
+        return cpu_limit_params.quota / cpu_limit_params.period if cpu_limit_params.quota != -1 else -1.0
+
+    def reset_cpu_limit(self) -> None:
+        self.controller_interface.set_cpu_limit_quota("-1")
+
+    def get_stat(self) -> Dict[str, int]:
+        stat_text = self.read_from_interface_file(self.controller_interface.cpu_stat)
+        return {line.split()[0]: int(line.split()[1]) for line in stat_text.splitlines()}
```
Jongy, line 110: Did you consider this class hierarchy (instead of your diamond hierarchy where D contains B C who extend A) - the methods here go to A, then B C will have them through inheritance, and just have a function D which gets the path and creates the appropriate object?

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+
+from granulate_utils.linux.cgroups.base_controller import BaseController
+from granulate_utils.linux.cgroups.cgroup import CgroupCore, CgroupCoreV1, CgroupCoreV2
+
+
+class MemoryControllerInterface:
+    memory_limit: str
+    usage: str
+    max_usage_in_bytes: str = ""
+    swap_limit: str = ""
```
Jongy, line 18: Why are these 2 initialized empty?

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+        pass
+
+    def get_memory_limit(self) -> int:
+        return int(self.cgroup.read_from_interface_file(self.memory_limit))
+
+    def _set_swap_limit(self, limit: str) -> None:
+        try:
+            self.cgroup.write_to_interface_file(self.swap_limit, limit)
+        except PermissionError:
+            # if swap extension is not enabled (CONFIG_MEMCG_SWAP) this file doesn't exist
```
Jongy, line 34:
```suggestion
            # if swap extension is not enabled (CONFIG_MEMCG_SWAP), this file doesn't exist
```

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+            # and PermissionError is thrown (since it can't be created)
```
Jongy, line 35: Btw instead of this thing you could just fix `write_to_interface_file` to use a Python API which writes to EXISTING files without trying to create them. The equivalent of NOT passing `O_CREAT`.

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+
+class MemoryControllerV1(MemoryControllerInterface):
+    memory_limit = "memory.limit_in_bytes"
+    swap_limit = "memory.memsw.limit_in_bytes"
+    max_usage_in_bytes = "memory.max_usage_in_bytes"
+    usage = "memory.usage_in_bytes"
+
+    def set_limit_in_bytes(self, limit: int) -> None:
+        # in case memsw.limit_in_bytes file exists we need to reset it in order to
+        # change limit_in_bytes in case it's smaller than memsw.limit_in_bytes
```
Jongy, line 47: Instead of temporarily removing the limit - just set the SAME limit here? Then you don't have to set it again afterwards (the second `_set_swap_limit` call)

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+    memory_limit = "memory.max"
+    swap_limit = "memory.swap.max"
+    usage = "memory.current"
+
+    def set_limit_in_bytes(self, limit: int) -> None:
+        # In CgroupV2 swap and memory are 2 different attributes that can be tuned, as opposed to Cgroup V1,
+        # in which we can just set a limit on both swap and memory together.
+        # So in Cgroup V2 we set swap limit to be 0.
+        swap_limit = "0"
+        memory_limit = str(limit)
```
Jongy, line 67: Code is more readable if you put these 2 assignments in an `else`. This makes it clear that EITHER these assignments or the `max` assignments are relevant

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+class MemoryControllerV2(MemoryControllerInterface):
+    UNBOUNDED_QUOTA_VALUE = "max"
+    memory_limit = "memory.max"
+    swap_limit = "memory.swap.max"
+    usage = "memory.current"
+
+    def set_limit_in_bytes(self, limit: int) -> None:
+        # In CgroupV2 swap and memory are 2 different attributes that can be tuned, as opposed to Cgroup V1,
+        # in which we can just set a limit on both swap and memory together.
+        # So in Cgroup V2 we set swap limit to be 0.
```
Jongy, line 65: If we keep it at 0, why do we change it at all? Just so that SOME limit is impsoed? Is it `max` by default?

granulate_utils/linux/cgroups/memory_controller.py
Jongy: Then you catch the `FileNotFoundError` and code is clearer

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+        super()._set_swap_limit("-1")
+        self.cgroup.write_to_interface_file(self.memory_limit, str(limit))
+
+        # memsw.limit_in_bytes is already set to -1
+        if limit != -1:
+            self._set_swap_limit(str(limit))
+
+
+class MemoryControllerV2(MemoryControllerInterface):
+    UNBOUNDED_QUOTA_VALUE = "max"
```
Jongy, line 57: Cosnt repeats - move to some global `CGROUP_V2_UNBOUNDED_VALUE`?

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+        # in which we can just set a limit on both swap and memory together.
+        # So in Cgroup V2 we set swap limit to be 0.
+        swap_limit = "0"
+        memory_limit = str(limit)
+        if limit == -1:
+            memory_limit = swap_limit = "max"
+        super()._set_swap_limit(swap_limit)
+        self.cgroup.write_to_interface_file(self.memory_limit, str(memory_limit))
+
+    def get_memory_limit(self) -> int:
```
Jongy, line 73: Docstring of parent class - `-1 is unlimited`

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+    swap_limit = "memory.swap.max"
+    usage = "memory.current"
+
+    def set_limit_in_bytes(self, limit: int) -> None:
+        # In CgroupV2 swap and memory are 2 different attributes that can be tuned, as opposed to Cgroup V1,
+        # in which we can just set a limit on both swap and memory together.
+        # So in Cgroup V2 we set swap limit to be 0.
+        swap_limit = "0"
+        memory_limit = str(limit)
+        if limit == -1:
```
Jongy, line 68: Docstring of parent class - `-1 is unlimited`

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(self, cgroup: CgroupCore):
+        self.cgroup = cgroup
+
+    @abstractmethod
+    def set_limit_in_bytes(self, limit: int) -> None:
+        pass
+
+    def get_memory_limit(self) -> int:
+        return int(self.cgroup.read_from_interface_file(self.memory_limit))
```
Jongy, line 28: This is only v1, isn't it? Move to v1 class

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def _set_swap_limit(self, limit: str) -> None:
```
Jongy, line 30: `limit: int`. I know that with `limit: str` you're able to reuse the same function for v1 & v2, but sry, it's a hack, the logic is different

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def get_memory_limit(self) -> int:
+        return self.controller_interface.get_memory_limit()
+
+    def get_usage_in_bytes(self) -> int:
+        return int(self.read_from_interface_file(self.controller_interface.usage))
+
+    def get_max_usage_in_bytes(self) -> int:
+        if self.controller_interface.max_usage_in_bytes == "":
+            raise Exception("Not implemented")
```
Jongy, line 98: If you go with my suggestion [here](https://github.com/Granulate/granulate-utils/pull/117/files#r1089821480) then this hack becomes nicely overriding `get_max_usage_in_bytes` in the v2 child, and raising an exception.

Anyway, wherever yopu end up putting the code, pls document why not implemented - I don't recall if v2 tracks this value at all? Is it supported? Did you just not implement here? Leave some breadcrumbs for the future developer adding it.

granulate_utils/linux/cgroups/memory_controller.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def get_memory_limit(self) -> int:
+        memory_limit = self.cgroup.read_from_interface_file(self.memory_limit)
+        if memory_limit == self.UNBOUNDED_QUOTA_VALUE:
+            return -1
+        return int(memory_limit)
+
+
+class MemoryController(BaseController):
+    controller = "memory"
```
Jongy, line 81: Btw the `.controller` class attr of `BaseController` should become `.CONTROLLER` IMO.

tests/granulate_utils/test_cgroups.py
```diff
@@ -1,28 +1,231 @@
...
+    cgroup_dir = Path(tmp_dir)
+    cpu_procs = cgroup_dir / "cgroup.procs"
+
+    cpu_procs.write_text("1 2 3")
+
+    cgroup_v1 = CgroupCoreV1(cgroup_dir)
+
+    assert cgroup_v1.get_pids_in_cgroup() == set([1, 2, 3])
+    cgroup_v1.assign_process_to_cgroup(4)
+    assert cgroup_v1.get_pids_in_cgroup() == set([4])  # write_text overwrites by default.
```
Jongy, line 30: `write_text` just overrides, not "by default"

```suggestion
    assert cgroup_v1.get_pids_in_cgroup() == set([4])  # write_text overwrites
```

tests/granulate_utils/test_cgroups.py
```diff
@@ -1,28 +1,231 @@
...
+
+    sub_cgroup_dir = cgroup_dir / "sub_cgroup"
+    assert not sub_cgroup_dir.exists()
+    sub_cgroup = cgroup_v1.get_cgroup_in_hierarchy("dummy", "sub_cgroup")
```
Jongy, line 34:
```suggestion
    sub_cgroup = cgroup_v1.get_cgroup_in_hierarchy("dummy_controller", "sub_cgroup")
```

Or better, add a global const `DUMMY_CONTROLLER`. Then, reading the test code makes it perfectly clear that the value is the same dummy used around, and that you don't care about it. Without it, I now tried to look around if the `dummy` value is used anywhere, before I recalled it's the `controller` name and it's not relevant for v1.

tests/granulate_utils/test_cgroups.py
```diff
@@ -1,28 +1,231 @@
...
+from pytest import TempPathFactory
+
+from granulate_utils.linux.cgroups.cgroup import CgroupCoreV1, CgroupCoreV2, get_cgroup_for_process
+from granulate_utils.linux.cgroups.cpu_controller import CpuController
+from granulate_utils.linux.cgroups.cpuacct_controller import CpuAcctController
+from granulate_utils.linux.cgroups.memory_controller import MemoryController
+
+
+# Cgroup
+def test_cgroup_sanity(tmp_path_factory: TempPathFactory):
```
Jongy, line 19:
```suggestion
def test_cgroup_v1(tmp_path_factory: TempPathFactory):
```
?

tests/granulate_utils/test_cgroups.py
```diff
@@ -1,28 +1,231 @@
...
+    tmp_dir = tmp_path_factory.mktemp("base_controller")
+    cgroup_dir = Path(tmp_dir)
+    cpu_procs = cgroup_dir / "cgroup.procs"
```
Jongy, line 22: Use the `cgroup.procs` const from your module

tests/granulate_utils/test_cgroups.py
Jongy, line 34: Consitfy `sub_cgroup` bc it's used more than once.

tests/granulate_utils/test_cgroups.py
```diff
@@ -1,28 +1,231 @@
...
+    assert enabled_controllers.read_text().strip() == f"+{CONTROLLER_TYPE}"
+
+    with pytest.raises(AssertionError) as exception:
+        sub_cgroup = cgroup.get_cgroup_in_hierarchy("dummy2", SUB_CGROUP_NAME)
+    assert exception.value.args[0] == "Controller not supported 'dummy2'"
+
+
+def test_get_cgroup_current_process():
+    root_path = Path("/root_path")
+    proc_fs_path = "/dummy"
```
Jongy, line 74: `proc_fs_path`? What's this name

tests/granulate_utils/test_cgroups.py
```diff
@@ -1,28 +1,231 @@
...
+            assert cgroup.path == full_path
+
+    with pytest.raises(Exception) as exception:
+        with patch("granulate_utils.linux.cgroups.cgroup.get_cgroup_mount", return_value=CgroupCoreV2(root_path)):
+            with patch(
+                "granulate_utils.linux.cgroups.cgroup.read_proc_file",
+                return_value=f"1:{CONTROLLER_TYPE}:/fail\n".encode(),
+            ):
+                cgroup = get_cgroup_for_process(CONTROLLER_TYPE)
+    assert exception.value.args[0] == f"'{CONTROLLER_TYPE}' not found"
```
Jongy, line 101: This fails because we found only v2 mount but the process has only a single v1 controller?

granulate_utils/linux/cgroups/base_controller.py
```diff
@@ -0,0 +1,54 @@
...
+from granulate_utils.linux.cgroups.cgroup import (
+    CgroupCore,
+    create_cgroup_from_path,
+    get_cgroup_for_process,
+    is_known_controller,
+)
+
+
+class BaseController:
+    controller: str  # class attribute to determine the controller type (should be initialized in inheriting classes)
```
Jongy, line 20: `CONTROLLER`

granulate_utils/linux/cgroups/base_controller.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        Assign process to this Cgroup
+        :param pid: pid of the process to assign (0 is current)
+        """
+        self.cgroup.assign_process_to_cgroup(pid)
+
+    def get_pids_in_cgroup(self) -> Set[int]:
+        return self.cgroup.get_pids_in_cgroup()
+
+    def read_from_interface_file(self, file_name: str) -> str:
```
Jongy, line 44:
```suggestion
    def read_from_interface_file(self, interface_name: str) -> str:
```

Consistency with the other class, fix the other methods as well

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -78,7 +101,94 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+        :param cgroup_name: the cgroup name
+        :return: bool indicating if cgroup is in hierarchy
+        """
+        return self._get_parent_cgroup(cgroup_name) is not None
+
+    def get_pids_in_cgroup(self) -> set[int]:
+        return {int(proc) for proc in self.read_from_interface_file(CGROUP_PROCS_FILE).split()}
+
+    def assign_process_to_cgroup(self, pid: int = 0) -> None:
+        self.write_to_interface_file(CGROUP_PROCS_FILE, str(pid))
```
Jongy, line 139: This is consistent behavior of assigning the PROCESS in both v1 and v2, right? Document for clarity with v1 differences and `tasks` file.

granulate_utils/linux/cgroups/base_controller.py
```diff
@@ -0,0 +1,54 @@
...
+        return self.cgroup.read_from_interface_file(file_name)
+
+    def write_to_interface_file(self, file_name: str, data: str) -> None:
+        self.cgroup.write_to_interface_file(file_name, data)
+
+    @classmethod
+    def get_cgroup_in_hierarchy(cls, cgroup_name: str, parent_cgroup: Optional[Union[Path, CgroupCore]] = None):
+        parent_controller = cls(parent_cgroup)
+        new_cgroup = parent_controller.cgroup.get_cgroup_in_hierarchy(cls.controller, cgroup_name)
+        return cls(new_cgroup)
```
Jongy, line 54: This method is weird.. Youget a `CgroupCore` object and create a Controller from which you take the CgroupCore to get another CgroupCore to create another controller? Gotta say I didn't follow. Try to simplify?

Jongy (dismissed):
Done reviewing

granulate_utils/linux/cgroups/cgroup.py
mirikl: This isn't an expected case but generally it could happen. Thinking about it, it's probably not that interesting and we  should just check the current cgroup and not the entire hierarchy.

granulate_utils/linux/cgroups/cgroup.py
mirikl: I removed it as I didn't find any usage of this function before this PR

granulate_utils/linux/cgroups/cgroup.py
mirikl: Yeah it looks correct to me:
`iter_mountinfo` will return `cpu` and `cpuacct` separated and then `find_v1_hierarchies` adds them as a separate key to the returned dictionary. However the path it returns will look like `/sys/fs/cgroup/cpu,cpuacct` but that's the way it works today.

granulate_utils/linux/cgroups/cpu_controller.py
mirikl: I'm adding `_INTERFACE` prefix instead of `_FILE`

granulate_utils/linux/cgroups/memory_controller.py
mirikl: Yeah, I did it so that when we set a limit on memory we won't just start using swap instead. Default is `max`.
As I understood the kernel docs, in V1 swap and memory had a limit on their sum, but in V2 these limits are separate, so it seemed logical to limit the swap to 0 when setting a limit on memory (to be consistent with V1 as far as we can), if you think of another solution LMK.
@michelhe

granulate_utils/linux/cgroups/memory_controller.py
mirikl: It's not implemented for v2. I can copy the docs from the kernel at the top of each controller to show what interfaces are supported and what they mean if you think it's helpful.

tests/granulate_utils/test_cgroups.py
mirikl: according to the glossary you suggested - `cgroup_path`

tests/granulate_utils/test_cgroups.py
mirikl: Yes

granulate_utils/linux/cgroups/base_controller.py
mirikl: I'll change the `get_cgroup_in_hierarchy` to `get_or_create_cgroup` and then this function just:
- Gets CgroupCore object to create the new Cgroup under
- Creates the new Cgroup
- Returns a controller which wraps the new Cgroup

granulate_utils/linux/cgroups/cpu_controller.py
mirikl: That's what I did at first actually.
I changed it mainly because of the function D, which needs to be duplicated in each controller type (cpu, memory...) and I didn't find a nice way to share this code between them.
The way it is now I only need `base_controller.get_cgroup_in_hierarchy` and the user just needs to call:
```
memory_controller = MemoryController.get_cgroup_in_hierarchy(cgroup_name='blabla')
```
Anyway, I think I like it better this way but I do see the benefit of the other way so if you want I don't mind changing.

granulate_utils/linux/cgroups/cgroup.py
mirikl: Specifically here, `cgroup_v2.is_controller_supported(controller)` is checking the controller is supported on the mount point of cgroup v2, which might be False if this controller isn't supported for v2 (i.e. the specific controller is not mounted for either v1 or v2)

mirikl: You are right that the usage of `is_controller_supported` in other contexts in the code, might return False because it's being used in a parent cgroup.

granulate_utils/linux/cgroups/cgroup.py
mirikl: I replied to one of your comments below, I think this is irrelevant and should be removed.

granulate_utils/linux/cgroups/cgroup.py
mirikl: The case you mentioned here shouldn't matter as we're looking at the root cgroup, our current process doesn't matter here.
Maybe it shouldn't be an assert and it should raise a `ControllerNotMounted` exception instead.

granulate_utils/linux/cgroups/memory_controller.py
mirikl: I moved the initialization inside the subclasses instead.

granulate_utils/linux/cgroups/cgroup.py
Jongy: The case I see is a bug if we re-create the same cgroup recursively.

Let's keep it simple and remove the parent logic - check just the current.

granulate_utils/linux/cgroups/cpu_controller.py
Jongy: The benefit is less indirections - no more classes containing classes, just classes extending classes, which is simpler to read & follow.

granulate_utils/linux/cgroups/memory_controller.py
mirikl: What you suggest might result in an error as the comment explains, I'll add more details to the comment though.
I found some docs about it [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/resource_management_guide/sec-memory) :
```
It is important to set the memory.limit_in_bytes parameter before setting the memory.memsw.limit_in_bytes parameter: attempting to do so in the reverse order results in an error. This is because memory.memsw.limit_in_bytes becomes available only after all memory limitations (previously set in memory.limit_in_bytes) are exhausted.
```
That means that if we don't reset the `memsw` limit it will result in an error:
1. If we set the `memsw` limit first - if we decrease it -> `memsw.limit < memory.limit`
2. If we set the `memory` limit first - if we increase it -> `memsw.limit < memory.limit`

granulate_utils/linux/cgroups/cgroup.py
mirikl: No, this is a bug. I should've tested that `cgroup.subtree_control` of the current cgroup has the controller we want. I'm changing it to raise an exception if the controller is not delegated as we can't create a subcgroup in this case.

granulate_utils/linux/cgroups/cgroup.py
mirikl: Discussed F2F, the solution we decided on was trying to create the new cGroup under the parent of current cGroup and if we cannot (e.g. the controller isn't delegated to the parent), we try to create it under the mount point

granulate_utils/linux/cgroups/cgroup.py
mirikl: https://github.com/Granulate/granulate-utils/pull/117#discussion_r1136654675 This comment is relevant here as well

granulate_utils/linux/cgroups/cgroup.py
d3dave: you can't remove it, because super options also contains fields which are not controller names

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -73,28 +95,230 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
-def get_cgroup_mount(controller: str, resolve_host_root_links: bool = True) -> Optional[str]:
+# Cgroup v1 and v2 handle the cgroup assigned processes using cgroup.procs
+CGROUP_PROCS_FILE = "cgroup.procs"
+
+
+class CgroupCore:
+    """
+    This object handles the general operations for Cgroup management:
+    hierarchy management, controllers management, assigning processes.
+    cgroup_full_path: The full path for this CGroupCore object
```
d3dave, line 110: > The full path for this CGroupCore object

what does this mean?

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -73,28 +95,230 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+            if parent_cgroup.is_controller_supported(controller):
+                return parent_cgroup
+        return CgroupCoreV2(self.cgroup_mount_path, self.cgroup_mount_path)
+
+    def get_subcgroup(self, controller: ControllerType, cgroup_name: str) -> CgroupCore:
+        """
+        Create a subcgroup. There are 2 cases:
+        1. Current cGroup is the root cGroup - in this case, we create our cGroup under current cGroup.
+        2. Current cGroup isn't the root cGroup - in this case, we create a cGroup under the parent of the
+           current cGroup.
```
d3dave, line 216: Then it's not a "sub-cgroup", is it? It would be a sibling. I think having the same function for both v1 and v2 makes this code harder to manage and follow. It should be clear that the logic we apply when in v1 is clearly creating a child cgroup, and when in v2 it should be clear that we create a sibling cgroup. Let's not hide it under one function with a vague name.

d3dave (changes requested):

granulate_utils/linux/cgroups/cgroup.py
mirikl: I changed the description a bit as it was too amorphic I think.
`cgroup_full_path` is `cgroup_absolute_path` - I guess I'll rename it.

granulate_utils/linux/cgroups/cgroup.py
mirikl: I agree with you, but this will kind of ruin the abstraction for the controllers, which will make the code much harder to use don't you think?

granulate_utils/linux/cgroups/cpu_controller.py
```diff
@@ -0,0 +1,119 @@
...
+from granulate_utils.linux.cgroups.cgroup import CgroupCore, ControllerType
+
+
+@dataclass
+class CpuLimitParams:
+    period: int
+    quota: int
+
+
+class CpuController(BaseController):
```
d3dave, line 23: it looks to me like we have methods to set cpu limit quota but no methods to set the cpu limit period. and i understand that in v2 both must always be set together. no reason to deprive ourselves of setting the period as well, the code should be simpler that way.

also, instead of setting each variable individually and creating a "special" case for v2, do the opposite; require setting always both with one function call, and "special" case v1.

d3dave (changes requested):

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -73,28 +95,231 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+        """
+        pass
+
+    @property
+    def is_v1(self) -> bool:
+        return False
+
+    @property
+    def is_v2(self) -> bool:
+        return False
```
d3dave, line 160: If you store the filesystem type in this class, then you don't need to ever override these properties:
```suggestion
    @property
    def is_v1(self) -> bool:
        return self.filesystem_type == 'cgroup'

    @property
    def is_v2(self) -> bool:
        return self.filesystem_type == 'cgroup2'
```

granulate_utils/linux/cgroups/cgroup.py
```diff
@@ -73,28 +95,231 @@ def find_v2_hierarchy(resolve_host_root_links: bool = True) -> Optional[str]:
...
+
+    for process_cgroup in get_process_cgroups(process):
+        if (cgroup_mount.is_v1 and controller in process_cgroup.controllers) or (
+            cgroup_mount.is_v2 and process_cgroup.hier_id == "0"
+        ):
+            # Validate hier == 0 only when cgroup is v2
+            assert cgroup_mount.is_v1 ^ (process_cgroup.hier_id == "0")
+            return cgroup_mount.build_object(
+                cgroup_mount.cgroup_abs_path / process_cgroup.relative_path.lstrip("/"), cgroup_mount.cgroup_mount_path
+            )
```
d3dave, line 310: Why do we need to check `hier_id == 0`?

d3dave (changes requested):

granulate_utils/linux/cgroups/cgroup.py
mirikl: I'd still need to override it in the subclaseses, do you prefer it because it's one property instead of 2?

granulate_utils/linux/cgroups/cgroup.py
mirikl: I removed the `assert`, but we need that check since in cgroupv2 the only indication we have that the cgroup line in `/proc/{pid}/cgroup` is the unified cgroup is that the `hier_id == 0`

d3dave (approved):

granulate_utils/linux/cgroups/base_controller.py
mirikl: Resolving as this is no longer relevant

granulate_utils/linux/cgroups/cgroup.py
mirikl: Resolving as this is no longer relevant

granulate_utils/linux/cgroups/cgroup.py
mirikl: Resolving as this is no longer relevant

granulate_utils/linux/cgroups/memory_controller.py
mirikl: Resolving as this is no longer relevant

PR #120, opened 2023-01-10
linux: handle exit()-ing process when reading auxv

d3dave:
Kernel frees task mm first before updating the process status, so it's state might be still "alive". But it will become zombie/dead very soon anyway.

### DIFF
--- a/granulate_utils/linux/process.py
+++ b/granulate_utils/linux/process.py
@@ -82,11 +82,11 @@ def read_process_execfn(process: psutil.Process) -> str:
 def _read_process_auxv(process: psutil.Process, auxv_id: int) -> int:
     auxv = read_proc_file(process, "auxv")
     if not auxv:
-        # kernel thread / zombie process yield this result.
-        # we don't expect to be called on kernel threads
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/process.py
```diff
@@ -82,11 +82,11 @@ def read_process_execfn(process: psutil.Process) -> str:
...
-        # kernel thread / zombie process yield this result.
-        # we don't expect to be called on kernel threads
-        assert not is_kernel_thread(process)
-        # so it's a zombie
-        assert is_process_zombie(process)
+        # Kernel threads and exit()-ed processes don't have auxv.
+        # We don't expect to be called on kernel threads
+        assert not is_kernel_thread(process), "attempted reading auxv of kthread!"
+        # The process status might still be alive until kernel updates it.
+        # That's ok, it will become zombie/dead very soon.
```
Jongy, line 89: You encountered cases where the zombie assert failed?

d3dave: Yes, on a process that became zombie soon after.

Jongy (approved):

PR #121, opened 2023-01-11
updated logging_kwargs template

kirshenboim:
## Use case
if you try to pass the following extra params : 
self.log(DEBUG, msg, *args, **kwargs) -> kwargs = {'name': 'pubsub', 'namespace_id': 2, 'hash': '123', 'old_hash': '567'}
you will get the following error:
"Attempt to overwrite name in LogRecord"
image

## Solution
if we would like to enable sending extra param keys from the following list
‘name’, ‘msg’, ‘args’, ‘levelname’, ‘levelno’, ‘pathname’, ‘filename’, ‘module’, ‘exc_info’, ‘exc_text’, ‘stack_info’, ‘lineno’, ‘funcName’, ‘created’, ‘msecs’, ‘relativeCreated’, ‘thread’, 
the extra param should not be at the same level of those logRecord keys

### DIFF
--- a/glogger/extra_adapter.py
+++ b/glogger/extra_adapter.py
@@ -39,6 +39,5 @@ class ExtraAdapter(logging.LoggerAdapter):
             logging_kwargs["extra"] = extra
 
         extra = self.get_extra(**logging_kwargs)
-        # Retain all extras as attributes on the record, and add "extra" attribute that contains all the extras:
-        logging_kwargs.update({"extra": {**extra, "extra": extra}})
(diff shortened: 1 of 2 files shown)


Jongy:
@d3dave please review this approach

d3dave:
Using generic names for extras makes querying harder. Prefer to use more specific names instead of just "name". If you really want this, then make it optional and preserve original behavior.

Jongy:
Closing, please reopen if you deem necessary & explain it

PR #122, opened 2023-01-30
Add jvm flags parsing

uriariel:
Add jvm flags parsing to use in gProfiler agent

### DIFF
--- a/granulate_utils/java.py
+++ b/granulate_utils/java.py
@@ -8,7 +8,7 @@ import re
 import signal
 from dataclasses import dataclass
 from itertools import dropwhile
-from typing import Iterable, List, Literal, Optional, Union
+from typing import Iterable, List, Literal, Match, Optional, Union
 
 from packaging.version import Version
 
@@ -215,3 +215,85 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
         zing_major = None
 
     return JvmVersion(version, build, vm_name, vm_type, zing_major)
+
+
+@dataclass
+class JvmFlag:
+    name: str
+    type: str
+    value: str
+    origin: str
+    kind: List[str]
+
+    @classmethod
+    def from_match(cls, match: Match) -> "JvmFlag":
+        # get the flag origin if jvm 9+, otherwise get is the flag from non default origin as described above
+        flag_origin_jdk_9 = match.group("flag_origin_jdk_9")
+
+        flag_equal_sign_prefix = match.group("flag_equal_sign_prefix")
+        flag_is_non_default_origin_only_jdk_8 = flag_equal_sign_prefix == ":"
+
+        is_jdk_8 = flag_is_non_default_origin_only_jdk_8 or flag_origin_jdk_9 is None
+
+        # split the list of space separated flag_kinds as described above
+        flag_kind = match.group("flag_kind").split()
+
+        if is_jdk_8:
+            if flag_is_non_default_origin_only_jdk_8:
+                flag_origin = "non-default"
+            else:
+                flag_origin = "default"
+        else:
+            flag_origin = flag_origin_jdk_9
+
+        return cls(
+            name=match.group("flag_name"),
+            type=match.group("flag_type"),
+            value=match.group("flag_value"),
+            origin=flag_origin,
+            kind=sorted(flag_kind),
+        )
+
+
+def parse_jvm_flags(jvm_flags_string: str) -> List[JvmFlag]:
+    """
+    The output of VM.flags -all format on jdk 8:
+    bool UseCompressedClassPointers               := true                                {lp64_product}
+    flag_type flag_name                           := flag_value                          {flag_kind}
+    ":=" indicates non default origin for the flag, while "=" indicates default origin
+
+    The output of VM.flags -all format on jdk 9+:
+    bool OptoScheduling                           = false                               {C2 pd product} {default}
+    flag_type flag_name                           = flag_value                          {flag_kind} {flag_origin}
+
+    flag_kind is space separated list of kinds, e.g. "C2 pd product"
+
+    possible flag kinds: "product", "manageable", "diagnostic", "experimental", "notproduct", "develop",
+    "lp64_product", "rw", "pd", "JVMCI", "C1", "C2", "ARCH"
+
+    possible flag types: "bool", "int", "uint", "intx", "uintx", "uint64_t", "size_t", "double", "ccstr",
+    "ccstrlist"
+
+    possible flag origins: default, non-default, command line, environment, config file, management, ergonomic,
+    attach, internal, jimage, "command line, ergonomic" (flag is set from command line and aligned by ergonomic)
+
+    KNOWN ISSUES:
+    - we don't parse the flag value for ccstrlist type flags (e.g. -XX:CompileCommand='A' -XX:CompileCommand='B')
+    """
+
+    vm_flags = []
+    vm_flags_pattern = re.compile(
+        r"(?P<flag_type>\S+)\s+"
+        r"(?P<flag_name>\S+)\s+"
+        r"(?P<flag_equal_sign_prefix>:)?= "
+        r"(?P<flag_value>\S+)\s*"
+        r"{(?P<flag_kind>.+?)}"
+        r"(?:\s*{(?P<flag_origin_jdk_9>default|non-default|command line|environment|config file|management|ergonomic|attach|internal|jimage|command line, ergonomic)})?"  # noqa: E501
+    )
+
+    for line in jvm_flags_string.splitlines():
+        match = vm_flags_pattern.search(line)
+        if match:
+            vm_flags.append(JvmFlag.from_match(match))
+
+    return vm_flags
--- a/granulate_utils/java.py
+++ b/granulate_utils/java.py
@@ -5,12 +5,13 @@
 
 from __future__ import annotations
 
+import dataclasses
 import os
 import re
 import signal
 from dataclasses import dataclass
 from itertools import dropwhile
-from typing import Iterable, List, Literal, Optional, Union
+from typing import Any, Dict, Iterable, List, Literal, Optional, Union
 
 from packaging.version import Version
 
@@ -235,6 +236,13 @@ class JvmFlag:
         r"{(?P<flag_kind>.+?)}"
         r"(?:\s*{(?P<flag_origin_jdk_9>.*)})?"
     )
+
+    def to_dict(self) -> Dict[str, Union[str, List[str]]]:
+        return dataclasses.asdict(self)
+
+    @classmethod
+    def from_dict(cls, jvm_flag_dict: Dict[str, Any]) -> JvmFlag:
+        return JvmFlag(**jvm_flag_dict)
 
     @classmethod
     def from_str(cls, line: str) -> Optional[JvmFlag]:
--- a/granulate_utils/java.py
+++ b/granulate_utils/java.py
@@ -231,7 +231,7 @@ class JvmFlag:
         r"(?P<flag_type>\S+)\s+"
         r"(?P<flag_name>\S+)\s+"
         r"(?P<flag_equal_sign_prefix>:)?= "
-        r"(?P<flag_value>\S+)\s+"  # We don't support empty string nor spaces in flag values, although its legal values
+        r"(?P<flag_value>\S+)\s+"  # noqa: E501 # We don't support empty string nor spaces in flag values, although both are legal values
         r"{(?P<flag_kind>.+?)}"
         r"(?:\s*{(?P<flag_origin_jdk_9>.*)})?"
     )
@@ -307,4 +307,7 @@ class JvmFlag:
 
 
 def parse_jvm_flags(jvm_flags_string: str) -> List[JvmFlag]:
-    return [flag for line in jvm_flags_string.splitlines() if (flag := JvmFlag.from_str(line)) is not None]
+    return sorted(
+        [flag for line in jvm_flags_string.splitlines() if (flag := JvmFlag.from_str(line)) is not None],
+        key=lambda flag: flag.name,
+    )
--- a/granulate_utils/java.py
+++ b/granulate_utils/java.py
@@ -5,10 +5,10 @@
 
 from __future__ import annotations
 
+import dataclasses
 import os
 import re
 import signal
-from dataclasses import dataclass
 from itertools import dropwhile
 from typing import Any, Dict, Iterable, List, Literal, Optional, Union
 
@@ -119,7 +119,7 @@ def java_exit_code_to_signo(exit_code: int) -> Optional[int]:
 VmType = Literal["HotSpot", "Zing", "OpenJ9", None]
 
 
-@dataclass
+@dataclasses.dataclass
 class JvmVersion:
     version: Version
     build: int
@@ -219,7 +219,7 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
     return JvmVersion(version, build, vm_name, vm_type, zing_major)
 
 
-@dataclass
+@dataclasses.dataclass
 class JvmFlag:
     name: str
     type: str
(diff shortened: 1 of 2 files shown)


granulate_utils/java.py
```diff
@@ -215,3 +215,85 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+        # get the flag origin if jvm 9+, otherwise get is the flag from non default origin as described above
+        flag_origin_jdk_9 = match.group("flag_origin_jdk_9")
+
+        flag_equal_sign_prefix = match.group("flag_equal_sign_prefix")
+        flag_is_non_default_origin_only_jdk_8 = flag_equal_sign_prefix == ":"
+
+        is_jdk_8 = flag_is_non_default_origin_only_jdk_8 or flag_origin_jdk_9 is None
+
+        # split the list of space separated flag_kinds as described above
+        flag_kind = match.group("flag_kind").split()
```
YoniKF, line 239: Please move these lines after the following if/else, so that all flag origin-related code is contiguous.

YoniKF (changes requested):

granulate_utils/java.py
```diff
@@ -215,3 +215,85 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+@dataclass
+class JvmFlag:
+    name: str
+    type: str
+    value: str
+    origin: str
+    kind: List[str]
+
+    @classmethod
+    def from_match(cls, match: Match) -> "JvmFlag":
```
Jongy, line 229: You can add `from __future__ import annotations` and then no `"JvmFlags"` but `JvmFlags`.

granulate_utils/java.py
```diff
@@ -215,3 +215,85 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+    bool OptoScheduling                           = false                               {C2 pd product} {default}
+    flag_type flag_name                           = flag_value                          {flag_kind} {flag_origin}
+
+    flag_kind is space separated list of kinds, e.g. "C2 pd product"
+
+    possible flag kinds: "product", "manageable", "diagnostic", "experimental", "notproduct", "develop",
+    "lp64_product", "rw", "pd", "JVMCI", "C1", "C2", "ARCH"
+
+    possible flag types: "bool", "int", "uint", "intx", "uintx", "uint64_t", "size_t", "double", "ccstr",
+    "ccstrlist"
```
Jongy, line 275: If you have a permalink to openjdk sources where these enums are defined, it'll be nice :)

Jongy: (for flag kinds & flag types)

granulate_utils/java.py
```diff
@@ -215,3 +215,85 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+    KNOWN ISSUES:
+    - we don't parse the flag value for ccstrlist type flags (e.g. -XX:CompileCommand='A' -XX:CompileCommand='B')
+    """
+
+    vm_flags = []
+    vm_flags_pattern = re.compile(
+        r"(?P<flag_type>\S+)\s+"
+        r"(?P<flag_name>\S+)\s+"
+        r"(?P<flag_equal_sign_prefix>:)?= "
+        r"(?P<flag_value>\S+)\s*"
```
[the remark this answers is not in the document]
Jongy: There seem to be spaces between the value & kind, in the output

granulate_utils/java.py
```diff
@@ -215,3 +215,85 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+        r"{(?P<flag_kind>.+?)}"
+        r"(?:\s*{(?P<flag_origin_jdk_9>default|non-default|command line|environment|config file|management|ergonomic|attach|internal|jimage|command line, ergonomic)})?"  # noqa: E501
+    )
+
+    for line in jvm_flags_string.splitlines():
+        match = vm_flags_pattern.search(line)
+        if match:
```
Jongy, line 296:
```suggestion
        if match is not None:
```

granulate_utils/java.py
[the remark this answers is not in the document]
Jongy: Agree :+1:

tests/granulate_utils/test_jvm_flags.py
```diff
@@ -0,0 +1,105 @@
...
+     intx NumberOfLoopInstrToAlign                 = 4                                      {C2 product} {management}
+     intx ObjectAlignmentInBytes                   = 8                                    {lp64_product} {internal}
+   size_t OldPLABSize                              = 1024                                 {ARCH product} {default}
+    uintx OldPLABWeight                            = 50                                        {product} {environment}
+   size_t OldSize                                  = 5452592                                   {product} {default}
+     bool OmitStackTraceInFastThrow                = true                                      {product} {attach}
+ccstrlist OnError                                  = cat hs_err_pid%p.log
+          OnError                                 += ps -ef                                    {product} {command line}
+ccstrlist OnOutOfMemoryError                       =                                     {C2 pd product} {default}
+     intx OnStackReplacePercentage                 = 140                                    {pd product} {config file}\n"""  # noqa: E501
```
Jongy, line 24: Why do you break the string here?

uriariel: long line, for the # noqa: E501

granulate_utils/java.py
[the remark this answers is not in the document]
uriariel: `ccstrlist OnOutOfMemoryError                       =                                     {C2 pd product} {default}`

its an example of legal flag with no value, and there is spaces between value and kind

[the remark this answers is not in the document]
uriariel: @YoniKF prefers to drop support for `ccstr` and `ccstrlist` since we cannot differentiate between `value=""` to `'value="   "`.
I think that its a weird case that we could document and keep support for `ccstr` and `ccstrlist`.
@Jongy wdyt?

granulate_utils/java.py
```diff
@@ -215,3 +215,85 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+    """
+
+    vm_flags = []
+    vm_flags_pattern = re.compile(
+        r"(?P<flag_type>\S+)\s+"
+        r"(?P<flag_name>\S+)\s+"
+        r"(?P<flag_equal_sign_prefix>:)?= "
+        r"(?P<flag_value>\S+)\s*"
+        r"{(?P<flag_kind>.+?)}"
+        r"(?:\s*{(?P<flag_origin_jdk_9>default|non-default|command line|environment|config file|management|ergonomic|attach|internal|jimage|command line, ergonomic)})?"  # noqa: E501
```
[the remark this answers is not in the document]
uriariel: I think its more robust, I would do it for kind but its a bit more complex, I can do it nevertheless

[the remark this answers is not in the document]
YoniKF: It is actually less robust because if there is a flag origin which is not specified, the parsing will fail. I think that accepting anything between the braces is fine, and if we want to filter out specific values we can use a blacklist instead of a whitelist.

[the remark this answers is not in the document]
uriariel: OK

granulate_utils/java.py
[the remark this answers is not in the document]
Jongy: I prefer to drop support until we solve it properly. Better be explicit with less support than wide support but implicit

granulate_utils/java.py
```diff
@@ -237,12 +236,13 @@ class JvmFlag:
         r"(?:\s*{(?P<flag_origin_jdk_9>.*)})?"
     )
 
-    def to_dict(self) -> Dict[str, Union[str, List[str]]]:
-        return dataclasses.asdict(self)
```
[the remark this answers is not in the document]
uriariel: But I already changed it to this format, as its easier and more efficient to parse in our backend, so we dont need to search for specific flags, but rather use `dict.get`

tests/granulate_utils/test_jvm_flags.py
```diff
@@ -49,13 +49,6 @@
                 JvmFlag(name="OldPLABWeight", type="uintx", value="50", origin="environment", kind=["product"]),
                 JvmFlag(name="OldSize", type="size_t", value="5452592", origin="default", kind=["product"]),
                 JvmFlag(name="OmitStackTraceInFastThrow", type="bool", value="true", origin="attach", kind=["product"]),
-                JvmFlag(
-                    name="OnOutOfMemoryError",
-                    type="ccstrlist",
```
YoniKF: You can maybe leave with `@xfail`?

granulate_utils/java.py
```diff
@@ -215,3 +217,94 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+    type: str
+    value: str
+    origin: str
+    kind: List[str]
+
+    vm_flags_pattern = re.compile(
+        r"(?P<flag_type>\S+)\s+"
+        r"(?P<flag_name>\S+)\s+"
+        r"(?P<flag_equal_sign_prefix>:)?= "
+        r"(?P<flag_value>\S+)\s+"  # We don't support empty string nor spaces in flag values, although its legal values
```
YoniKF, line 234: Add to the known issues.

YoniKF, line 234:
```suggestion
        r"(?P<flag_value>\S+)\s+"  # We don't support empty string nor spaces in flag values, although both are legal values
```

granulate_utils/java.py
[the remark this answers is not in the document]
YoniKF: You can keep the format of a dict name->attributes, but the current `to_dict`/`from_dict` methods are unnatural: `from_dict` receives a dict, but assumes that it only has a single key; `from_dict` returns a dict, but it's callers will have to unpack it to a pair of name and attributes to be able to use it.

granulate_utils/java.py
```diff
@@ -215,3 +217,94 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+        possible flag kinds:
+        "product", "manageable", "diagnostic", "experimental", "notproduct", "develop", "lp64_product", "rw", "pd", "JVMCI", "C1", "C2", "ARCH"
+        https://github.com/openjdk/jdk17u/blob/2fe42855c48c49b515b97312ce64a5a8ef3af407/src/hotspot/share/runtime/flags/jvmFlag.cpp#L338 # noqa: E501
+
+        possible flag types:
+        "bool", "int", "uint", "intx", "uintx", "uint64_t", "size_t", "double", "ccstr", "ccstrlist"
+        https://github.com/openjdk/jdk17u/blob/2fe42855c48c49b515b97312ce64a5a8ef3af407/src/hotspot/share/runtime/flags/jvmFlag.hpp#L134 # noqa: E501
+
+        possible flag origins:
+        default, non-default, command line, environment, config file, management, ergonomic, attach, internal, jimage, "command line, ergonomic" (flag is set from command line and aligned by ergonomic) # noqa: E501
```
Jongy, line 270: "command line, ergonomic" doesn't appear in that enum o.O

tests/granulate_utils/test_jvm_flags.py
```diff
@@ -0,0 +1,131 @@
...
+)
+def test_parse_jvm_flags(jvm_flags_string: str, expected_jvm_flags_list: List[JvmFlag]) -> None:
+    assert parse_jvm_flags(jvm_flags_string) == expected_jvm_flags_list
+
+
+@pytest.mark.parametrize(
+    "jvm_flag,expected_jvm_flag_serialized",
+    [
+        (
+            JvmFlag(
```
Jongy, line 112: Add something here that proves the sorting during serialization? (or parsing)

Jongy: i.e a test case

Jongy:
@YoniKF @uriariel good to merge once my comments are resolved

granulate_utils/java.py
uriariel: https://github.com/openjdk/jdk17u/blob/2fe42855c48c49b515b97312ce64a5a8ef3af407/src/hotspot/share/runtime/flags/jvmFlag.cpp#L397

granulate_utils/java.py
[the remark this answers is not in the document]
uriariel: but we want to keep serialization/deserialization in one place, how you suggest to do it?

tests/granulate_utils/test_jvm_flags.py
uriariel: serialization/deserialization is for a single flag, and parsing is not guaranteeing alphabetical sort, but a sorted list according to the input string (which is tested in the sanity case).

fixed the parsing to return a sorted list, as we were accounting on ordered `VM.flags` output string.

granulate_utils/java.py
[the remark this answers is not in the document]
YoniKF: Use `dataclasses.asdict`, serialize all the flags to a list of dicts.
The code that is interested in a specific flag will have to either construct a dict name->JvmFlag or just search for the required flag.

tests/granulate_utils/test_jvm_flags.py
```diff
@@ -105,6 +105,35 @@ def test_parse_jvm_flags(jvm_flags_string: str, expected_jvm_flags_list: List[Jv
...
+    [
+        (
+            """
+ccstrlist OnError                                  = echo a
+          OnError                                 += echo b                                    {product} {command line}
+ccstrlist OnOutOfMemoryError                       =                                     {C2 pd product} {default}
+""",
+            [
+                JvmFlag(
+                    name="OnError", type="ccstrlist", value="echo a, echo b", origin="command line", kind=["product"]
```
YoniKF, line 120: Is the expected value `"echo a, echo b"` or `"echo a\necho b"`?

granulate_utils/java.py
YoniKF: I don't see it under the known issues.

YoniKF (changes requested):

tests/granulate_utils/test_jvm_flags.py
YoniKF: The sorting is necessary when sending the flags as metadata, i.e., it should appear in the profiler code after filtering out the undesired flags. This module in granulate-utils only parses the output of `jcmd VM.flags` - it isn't under its responsibility to change their order.

Jongy: You are calling `sorted` in this code, so my suggestion is to incorporate a test that ensures sorting remains in effect.

granulate_utils/java.py
Jongy: Gotcha

tests/granulate_utils/test_jvm_flags.py
uriariel: I called it because of the test

tests/granulate_utils/test_jvm_flags.py
YoniKF: Actually, if we add support it will have to be `["echo a", "echo b"]`.

granulate_utils/java.py
```diff
@@ -215,3 +217,97 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+            type=match.group("flag_type"),
+            value=match.group("flag_value"),
+            origin=flag_origin,
+            kind=sorted(flag_kind),
+        )
+
+
+def parse_jvm_flags(jvm_flags_string: str) -> List[JvmFlag]:
+    return sorted(
+        [flag for line in jvm_flags_string.splitlines() if (flag := JvmFlag.from_str(line)) is not None],
```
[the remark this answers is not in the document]
uriariel: removed the sort

tests/granulate_utils/test_jvm_flags.py
Jongy: k

granulate_utils/java.py
```diff
@@ -215,3 +217,94 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+        r"{(?P<flag_kind>.+?)}"
+        r"(?:\s*{(?P<flag_origin_jdk_9>.*)})?"
+    )
+
+    def to_dict(self) -> Dict[str, Union[str, List[str]]]:
+        return dataclasses.asdict(self)
+
+    @classmethod
+    def from_dict(cls, jvm_flag_dict: Dict[str, Any]) -> JvmFlag:
+        return cls(**jvm_flag_dict)
```
YoniKF, line 244: Consider removing the method and just using `JvmFlag(...)` anywhere you use `JvmFlag.from_dict(...)`.

YoniKF (approved):
One open comment, decide for yourself if you want to do it or not (if not you can resolve it and merge).

granulate_utils/java.py
uriariel: I prefer to leave it and `to_dict` for ease of use and clarity

PR #124, opened 2023-02-23
Change Python process regex to ERE friendly

mpozniak95:
As mentioned in https://github.com/Granulate/gprofiler/pull/701#discussion_r1115640037

### DIFF
--- a/granulate_utils/python.py
+++ b/granulate_utils/python.py
@@ -3,6 +3,6 @@
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
 
-DETECTED_PYTHON_PROCESSES_REGEX = r"(?:^.+/(?:lib)?python[^/]*$)|(?:^.+/site-packages/.+?$)|(?:^.+/dist-packages/.+?$)"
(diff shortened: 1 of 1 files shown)


Jongy:
@mpozniak95 can you provide some reference to this change? ("ERE friendly"?). Did you test it on other systems to ensure correctness, i.e it will still work on Debian/CentOS `grep`?

mpozniak95:
According to: https://linuxcommand.org/lc3_man_pages/grep1.html. Previously we used grep with -P which is:

> Interprets PATTERNS as Perl-compatible  regular  expressions (PCREs)

since busybox is lacking PCRE, grep package on Alpine doesn't support -P flag. This is the reason why I changed from -P to -E which is:

> Interprets  PATTERNS  as extended regular expressions EREs

Since tests on my PR in gprofiler passed with that changes: https://github.com/Granulate/gprofiler/actions/runs/4252348467/jobs/7396272000 I think it doesn't break anything and it will work on Debian/CentOS

Jongy:
Understood :)

PR #125, opened 2023-02-27
glogger: Allow postponed initialization of Sender

mirikl:


netaneld122 (dismissed):
LGTM

DanielShaulov (dismissed):

d3dave (commented):
Please update the docstring accordingly.

netaneld122 (approved):
LGTM, I want @DanielShaulov to have a look as well

glogger/handler.py
```diff
@@ -51,21 +51,31 @@ def __init__(
...
 
         self.stdout_logger = get_stdout_logger()
         self.jsonify = JSONEncoder(separators=(",", ":"), default=repr).encode  # compact, no whitespace
         self.messages_buffer = MessagesBuffer(max_total_length, overflow_drop_factor)
         self.messages_buffer.head_serial_no = continue_from
 
+        self.sender: Optional[Sender] = None
+        self.init_sender(sender)
+
+    def init_sender(self, sender: Optional[Sender]) -> None:
```
netaneld122, line 64: Actually, let's accept a non-optional `Sender` here and check if it's None in the __init__ function instead

DanielShaulov:
What happened here? What was wrong with approved changes for Yesterday?

mirikl:
> What happened here? What was wrong with approved changes for Yesterday?

@DanielShaulov , after talking with @netaneld122 it seems that this change is cleaner then the other one.
The change in sender is more complicated because we need to perform the change we did for `auth_token` to `server_address` also.

DanielShaulov (approved):

PR #126, opened 2023-02-28
Migrate metrics code from gprofiler

d3dave:
Mostly deals with YARN metrics collection as in gProfiler. Migrated the YARN+Spark URL paths and metrics names, as well as the YARN collection logic. Will add Spark collection after a bit more cleanup.

### DIFF
--- a/granulate_utils/metrics/__init__.py
+++ b/granulate_utils/metrics/__init__.py
@@ -1,4 +1,41 @@
 from typing import Any, Dict
+from urllib.parse import urljoin
+
+import requests
+
+
+def rest_request_to_json(address: str, object_path: str, *args: Any, **kwargs: Any) -> Any:
+    """
+    Query the given URL and return the JSON response
+    """
+    return rest_request(address, object_path, *args, **kwargs).json()
+
+
+def rest_request(url: str, object_path: str, *args: Any, **kwargs: Any) -> requests.Response:
+    """
+    Query the given URL and return the response
+    """
+    if object_path:
+        url = join_url_dir(url, object_path)
+
+    # Add args to the url
+    if args:
+        for directory in args:
+            url = join_url_dir(url, directory)
+
+    response = requests.get(url, params={k: v for k, v in kwargs.items() if v is not None}, timeout=3)
+    response.raise_for_status()
+    return response
+
+
+def join_url_dir(url: str, *args: Any) -> str:
+    """
+    Join a URL with multiple directories
+    """
+    for path in args:
+        url = url.rstrip("/") + "/"
+        url = urljoin(url, path.lstrip("/"))
+    return url
 
 
 def set_individual_metric(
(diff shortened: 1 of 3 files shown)


granulate_utils/metrics/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    if object_path:
+        url = join_url_dir(url, object_path)
+
+    # Add args to the url
+    if args:
+        for directory in args:
+            url = join_url_dir(url, directory)
+
+    response = requests.get(url, params={k: v for k, v in kwargs.items() if v is not None}, timeout=3)
```
Jongy, line 33:
```suggestion
    response = requests.get(url, params=dict(kwargs or {}), timeout=3)
```
?

granulate_utils/metrics/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def set_individual_metric(
+    collected_metrics: Dict[str, Dict[str, Any]], name: str, value: Any, labels: Dict[str, str]
+) -> None:
+    """
+    Add a metric to collected_metrics with labels in {name, value, labels} format.
+    Metric is only added if the value is not None.
+    """
+    if name not in collected_metrics and value is not None:
```
Jongy, line 55: Legacy code but - do you know WHY we check if metric already exists? i.e in which flow this can happen that it already exists?

granulate_utils/metrics/spark.py
```diff
@@ -0,0 +1,92 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+# (C) Datadog, Inc. 2018-present. All rights reserved.
+# Licensed under a 3-clause BSD style license (see LICENSE.bsd3).
+#
+SPARK_APPS_PATH = "api/v1/applications"
+
+SPARK_JOB_METRICS = {
```
Jongy, line 10: Nice rename catch :)

Jongy (changes requested):
Done reviewing, small

granulate_utils/metrics/__init__.py
d3dave: Not the same effect.

granulate_utils/metrics/__init__.py
d3dave: Should not happen afaik. Should I add an assert?

Jongy: Yes

granulate_utils/metrics/__init__.py
```diff
@@ -52,7 +52,9 @@ def set_individual_metric(
     Add a metric to collected_metrics with labels in {name, value, labels} format.
     Metric is only added if the value is not None.
     """
-    if name not in collected_metrics and value is not None:
+    assert name not in collected_metrics, "attempted to add same metric twice!"
```
Jongy, line 55: give the metric name in assert message?

Jongy (changes requested):
1 small

Jongy (approved):

PR #127, opened 2023-03-06
metrics: yarn: ResourceManager API

d3dave:
Add a `ResourceManager` API class that can be used independently of the `YarnCollector`.


Jongy (approved):
@d3dave please refrain from force pushes next times :pray:

PR #128, opened 2023-03-06
metrics: API client

d3dave:
Add common skeleton for API client implementation. Users should provide their own definition for `headers()`, and `log_*()`.

### DIFF
--- a/granulate_utils/metrics/api_client.py
+++ b/granulate_utils/metrics/api_client.py
@@ -87,10 +87,10 @@ class APIClient:
     def log_response(self, response: requests.Response) -> None:
         pass
 
-    def submit_spark_metrics(self, url: str, timestamp: int, metrics: List[Dict[str, Any]]) -> Dict:
+    def submit_spark_metrics(self, path: str, timestamp: int, metrics: List[Dict[str, Any]]) -> Dict:
         return self._request_url(
             "POST",
-            url,
+            f"{self._server_address}/{path}",
             {
                 "format_version": 0,
                 "timestamp": timestamp,
--- a/granulate_utils/metrics/api_client.py
+++ b/granulate_utils/metrics/api_client.py
@@ -1,0 +1,100 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+import gzip
+import json
+from io import BytesIO
+from typing import IO, Any, Dict, List, cast
+
+import requests
+
+DEFAULT_API_SERVER_ADDRESS = "https://api.granulate.io"
+DEFAULT_REQUEST_TIMEOUT = 5
+DEFAULT_UPLOAD_TIMEOUT = 120
+
+
+class APIError(Exception):
+    def __init__(self, message: str, full_data: dict = None):
+        self.message = message
+        self.full_data = full_data
+
+    def __str__(self) -> str:
+        return self.message
+
+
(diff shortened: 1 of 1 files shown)


granulate_utils/metrics/api_client.py
```diff
@@ -0,0 +1,98 @@
...
+#
+import gzip
+import json
+from io import BytesIO
+from typing import IO, Any, Dict, List, Optional, cast
+
+import requests
+
+DEFAULT_API_SERVER_ADDRESS = "https://api.granulate.io"
+DEFAULT_REQUEST_TIMEOUT = 5
```
Jongy, line 13: These consts are taken from what we currently use in gProfiler?

granulate_utils/metrics/api_client.py
```diff
@@ -0,0 +1,98 @@
...
+DEFAULT_UPLOAD_TIMEOUT = 120
+
+
+class APIError(Exception):
+    def __init__(self, message: str, full_data: dict = None):
```
Jongy, line 18: IIRC `dict` as type is 3.9 and this code needs to support 3.8 as well

granulate_utils/metrics/api_client.py
```diff
@@ -0,0 +1,98 @@
...
+    def headers(self) -> Dict[str, str]:
+        return {}
+
+    def log_bad_json(self, data: Any) -> None:
+        pass
+
+    def log_response(self, response: requests.Response) -> None:
+        pass
+
+    def submit_spark_metrics(self, path: str, timestamp: int, metrics: List[Dict[str, Any]]) -> Dict:
```
Jongy, line 88: Note that we plan to make `timestamp` a string later. We'll change it later then?

granulate_utils/metrics/api_client.py
```diff
@@ -0,0 +1,98 @@
...
+class APIError(Exception):
+    def __init__(self, message: str, full_data: dict = None):
+        self.message = message
+        self.full_data = full_data
+
+    def __str__(self) -> str:
+        return self.message
+
+
+class APIClient:
```
Jongy, line 26: This replaces gProfiler's `BaseAPIClient`  or currently just a copy?

Jongy: I see you built it perfectly and gProfiler will be able to extend it with `log_bad_json` and `log_response`. Good job

granulate_utils/metrics/api_client.py
Jongy: Yes, checked, this is what we have in `client.py`.

granulate_utils/metrics/api_client.py
```diff
@@ -0,0 +1,98 @@
...
+        **kwargs,
+    ) -> Dict:
+        headers = self.headers().copy()
+        if params is None:
+            params = {}
+
+        if method.upper() == "GET":
+            if data is not None:
+                params.update(data)
+                data = None
```
Jongy, line 52: Small possible fix for the future, nice

granulate_utils/metrics/api_client.py
Jongy, line 88: Can this be in another class? `ProfilerAPIClient` will extend this class I suppose, and it doesn't need `submit_spark_metrics`.

Jongy (changes requested):
Small

Jongy:
@d3dave not needed anymore I think? Can you close?

PR #129, opened 2023-03-08
containers: Add create_time & start_time for containers

michelhe:


granulate_utils/containers/docker.py
```diff
@@ -35,8 +36,23 @@ def get_runtimes(self) -> List[str]:
...
+    def _parse_docker_timestamp(time_str: str) -> Optional[datetime]:
+        """
+        Parses timestamps provided by DockerAPI to datetime.
+        Docker works with ISO format timestamps (in UTC) with fractional milliseconds that python standard library
+        doesn't parse, and also ends with "Z" timezone indicator for UTC.
+        """
+        assert time_str.endswith('Z')  # assert UTC
+        if time_str.startswith('0001'):
+            return None
+        return datetime.fromisoformat(time_str.split('.')[0])
```
d3dave, line 48: Have you tried with `dateutil`?

d3dave (changes requested):

granulate_utils/containers/docker.py
Jongy: dateutil will succeed here indeed

michelhe: Will probably work, but I didn't want to introduce another requirement

granulate_utils/containers/cri.py
```diff
@@ -73,49 +74,56 @@ def list_containers(self, all_info: bool) -> List[Container]:
...
     def get_runtimes(self) -> List[str]:
         return list(self._runtimes.keys())
 
     @classmethod
-    def _create_container(
-        cls, container: Union[api_pb2.Container, api_pb2.ContainerStatus], pid: Optional[int], runtime: str
+    def _create_container_from_status(
+        cls, status: api_pb2.ContainerStatus, pid: Optional[int], runtime: str
     ) -> Container:
+        created_at_ns = cast(int, status.created_at)
```
Jongy, line 113: Use `assert_case` in such sites

granulate_utils/containers/cri.py
```diff
@@ -73,49 +74,56 @@ def list_containers(self, all_info: bool) -> List[Container]:
...
-                        status = stub.ContainerStatus(
-                            api_pb2.ContainerStatusRequest(container_id=container.id, verbose=True)
-                        )
-                        pid: Optional[int] = json.loads(status.info.get("info", "{}")).get("pid")
-                    else:
-                        pid = None
-
-                    containers.append(self._create_container(container, pid, rt))
+                for c in stub.ListContainers(api_pb2.ListContainersRequest()).containers:
+                    container = self._get_container_from_runtime(c.id, rt, stub, all_info)
```
Jongy, line 78: You are ignoring `all_info` here. Do not call `ContainerStatus` if you don't need

Jongy (changes requested):

granulate_utils/containers/cri.py
michelhe: The `get_container` API always calls `ContainerStatus` and  uses `all_info` to control the verbose filed of the `ContainerStatusRequest`.
I need the start_time to always be available when I list containers, otherwise I need some way to differntiate a tri-state in the container dataclass for that field
1. start_time=Some(time) 
2. start_time=None (container has not started yet)
3. start_time=<MISSING>  - didn't pass `all_info`, I don't know if started or not

Jongy: You want the time for whatever usage when **you** list containers, but when gProfiler lists, it doesn't require, so don't add it in this API, add it in another API or expose it with a parameter that keeps the old behavior default. You're adding N API calls...

granulate_utils/containers/cri.py
Jongy: hmm it's in gprofiler, u can move or meh

Jongy: hmm it's in gprofiler, u can move or meh

granulate_utils/containers/docker.py
michelhe: Yes but its redundant

d3dave: > Yes but its redundant

Why is it redundant?

d3dave (dismissed):

granulate_utils/containers/container.py
```diff
@@ -4,9 +4,16 @@
...
 
 from dataclasses import dataclass
+from datetime import datetime
 from typing import Dict, List, Optional
 
 
+@dataclass
+class TimeInfo:
+    create_time: datetime  # Creation time of the container (UTC)
+    start_time: Optional[datetime]  # Start time of the container (UTC) - None=not started
```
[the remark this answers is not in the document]
michelhe: If under container, then I have a tri-state problem where None values can mean either container not started or data wasn't queried.

With this approach, None time_info means data wasn't queried and None start time means container is stopped

granulate_utils/containers/docker.py
michelhe: Because the user knows its UTC because this is what I wrote in the comment next to the class var definition :)

d3dave: any operation on a datetime object that does not contain timezone info assumes its in local time. it does not matter what the user knows, the code in the datetime module doesnt

granulate_utils/containers/docker.py
```diff
@@ -34,16 +35,21 @@ def get_container(self, container_id: str, all_info: bool) -> Container:
...
         return ["docker"]
 
-    @staticmethod
-    def _create_container(container: docker.models.containers.Container) -> Container:
+    @classmethod
+    def _create_container(cls, container: docker.models.containers.Container) -> Container:
         pid: Optional[int] = container.attrs["State"].get("Pid")
+        created = isoparse(container.attrs["Created"])
+        assert created is not None
+        started_at = isoparse(container.attrs["State"]["StartedAt"])
```
Jongy, line 43: If not started yet?

granulate_utils/containers/cri.py
```diff
@@ -75,32 +77,37 @@ def list_containers(self, all_info: bool) -> List[Container]:
...
                     else:
-                        pid = None
-
-                    containers.append(self._create_container(container, pid, rt))
+                        containers.append(self._create_container(container, rt, None))
 
         return containers
 
+    def _get_container_status(
+        self, stub: RuntimeServiceStub, container: Union[api_pb2.Container, str], *, verbose: bool
```
Jongy, line 90: Any reason to accept a Union and not have the callers per site do it properly?

granulate_utils/containers/cri.py
```diff
@@ -75,32 +77,37 @@ def list_containers(self, all_info: bool) -> List[Container]:
             with RuntimeServiceWrapper(path) as stub:
                 for container in stub.ListContainers(api_pb2.ListContainersRequest()).containers:
                     if all_info:
-                        # need verbose=True to get the info which contains the PID
-                        status = stub.ContainerStatus(
-                            api_pb2.ContainerStatusRequest(container_id=container.id, verbose=True)
-                        )
-                        pid: Optional[int] = json.loads(status.info.get("info", "{}")).get("pid")
+                        status_response = self._get_container_status(stub, container, verbose=all_info)
+                        assert status_response is not None, "container went down"
```
Jongy, line 81: Why assert? If you're fixing the bug here then handle it properly lol. `continue`

granulate_utils/containers/cri.py
```diff
@@ -75,32 +77,37 @@ def list_containers(self, all_info: bool) -> List[Container]:
             with RuntimeServiceWrapper(path) as stub:
                 for container in stub.ListContainers(api_pb2.ListContainersRequest()).containers:
                     if all_info:
-                        # need verbose=True to get the info which contains the PID
-                        status = stub.ContainerStatus(
-                            api_pb2.ContainerStatusRequest(container_id=container.id, verbose=True)
-                        )
-                        pid: Optional[int] = json.loads(status.info.get("info", "{}")).get("pid")
+                        status_response = self._get_container_status(stub, container, verbose=all_info)
```
Jongy, line 80:
```suggestion
                        status_response = self._get_container_status(stub, container, verbose=True)
```
and retain the comment please

Jongy (changes requested):
Done reviewing

granulate_utils/containers/docker.py
michelhe: Oh fuck I had logic to handle it here [73a00e4](https://github.com/Granulate/granulate-utils/pull/129/commits/73a00e4e560936979a7f53752fd693c3966125f8)

granulate_utils/containers/docker.py
```diff
@@ -35,15 +37,27 @@ def get_runtimes(self) -> List[str]:
         return ["docker"]
 
     @staticmethod
-    def _create_container(container: docker.models.containers.Container) -> Container:
+    def _parse_docker_ts(ts: str) -> Optional[datetime]:
+        assert ts.endswith("Z")  # assert UTC
+        if ts.startswith("0001"):
```
Jongy, line 42: What? You're expecting some magic value here, please comment it here

granulate_utils/containers/docker.py
```diff
@@ -35,15 +37,27 @@ def get_runtimes(self) -> List[str]:
...
+            return None
+        return isoparse(ts)
+
+    @classmethod
+    def _create_container(cls, container: docker.models.containers.Container) -> Container:
         pid: Optional[int] = container.attrs["State"].get("Pid")
+        created = cls._parse_docker_ts(container.attrs["Created"])
+        assert created is not None
+        started_at = cls._parse_docker_ts(container.attrs["State"]["StartedAt"])
```
Jongy, line 51: The key `"StartedAt" is always present?

granulate_utils/containers/cri.py
Jongy: @michelhe you didn't do it, please retain the comment `# need verbose=True to get the info which contains the PID` and change to `verbose=True` for clarity.

granulate_utils/containers/docker.py
```diff
@@ -35,15 +37,27 @@ def get_runtimes(self) -> List[str]:
...
         if pid == 0:  # Docker returns 0 for dead containers
             pid = None
```
Jongy, line 53: Move this code block next to `pid` handling, so there are 3 lines handling `pid` then the lines handling the times stuff and then the `Container` object

granulate_utils/containers/cri.py
michelhe: Did do, forgot to push sorry

granulate_utils/containers/cri.py
```diff
@@ -75,32 +77,35 @@ def list_containers(self, all_info: bool) -> List[Container]:
...
-                    )
-                except grpc._channel._InactiveRpcError as e:
-                    if e.code() == grpc.StatusCode.NOT_FOUND:
-                        continue
-                    raise
-
-                pid: Optional[int] = json.loads(status.info.get("info", "{}")).get("pid")
-                return self._create_container(status.status, pid, rt)
+                status_response = self._container_status_request(stub, container_id, verbose=all_info)
+                if not status_response:
```
Jongy, line 105:
```suggestion
                if status_response is None:
```

granulate_utils/containers/cri.py
```diff
@@ -75,32 +77,35 @@ def list_containers(self, all_info: bool) -> List[Container]:
             with RuntimeServiceWrapper(path) as stub:
                 for container in stub.ListContainers(api_pb2.ListContainersRequest()).containers:
                     if all_info:
-                        # need verbose=True to get the info which contains the PID
-                        status = stub.ContainerStatus(
-                            api_pb2.ContainerStatusRequest(container_id=container.id, verbose=True)
-                        )
-                        pid: Optional[int] = json.loads(status.info.get("info", "{}")).get("pid")
+                        status_response = self._container_status_request(stub, container.id, verbose=all_info)
+                        if not status_response:
```
Jongy, line 81:
```suggestion
                        if status_response is None:
```

granulate_utils/containers/cri.py
```diff
@@ -109,13 +114,27 @@ def get_runtimes(self) -> List[str]:
 
     @classmethod
     def _create_container(
-        cls, container: Union[api_pb2.Container, api_pb2.ContainerStatus], pid: Optional[int], runtime: str
```
Jongy, line 112: Revert the API change? you're still passing `pid` every time

Jongy (changes requested):
Done re-reviewing

granulate_utils/containers/docker.py
michelhe: Added a comment, but in docker engine the timestamps are non-null strings so the zero value is 0001-01-... timestamp

granulate_utils/containers/docker.py
michelhe: Yes, its part of the `ContainerState` type in the docker engine

Jongy (approved):

PR #130, opened 2023-03-16
granulate_utils: Improve cloud info fetching performance

netaneld122:
* Add the `call_in_parallel` utility for convenience
* Use `call_in_parallel` in `get_static_cloud_instance_metadata` so we don't waste time if we're in the wrong cloud provider

### DIFF
--- a/granulate_utils/futures.py
+++ b/granulate_utils/futures.py
@@ -1,0 +1,29 @@
+from concurrent.futures import ThreadPoolExecutor, as_completed, _base, Future
+from contextlib import contextmanager
+from typing import Callable, Collection, Generator, TypeVar
+
+T = TypeVar('T')
+
+
+@contextmanager
+def wrap_thread_pool(pool: ThreadPoolExecutor) -> Generator[ThreadPoolExecutor, None, None]:
+    try:
+        yield pool
+    except _base.TimeoutError as e:
+        raise TimeoutError(*e.args) from e
+    finally:
+        pool.shutdown(wait=False, cancel_futures=True)
+
+
+def call_in_parallel(callables: Collection[Callable[[], T]], timeout: float, max_threads: int = 10) -> Generator[Future[T], None, None]:
+    """
+    Call the given callables in parallel and generate their futures in correspondence to their finishing order.
+
+    Please note that the underlying threads are not daemonized and this function does not wait for all calls to finish,
+    this means that if the caller does not invoke future.result() for all of the calls - the other callables will continue executing in the background (potentially preventing the process from closing).
+    Therefore it's important to make sure the callables themselves have timeouts.
+    """
+    with wrap_thread_pool(ThreadPoolExecutor(max_workers=min(len(callables), max_threads))) as pool:
+        futures = {pool.submit(callable) for callable in callables}
+        for completed in as_completed(futures, timeout):
+            yield completed
--- a/granulate_utils/futures.py
+++ b/granulate_utils/futures.py
@@ -1,8 +1,6 @@
 from concurrent.futures import Future, ThreadPoolExecutor, _base, as_completed
 from contextlib import contextmanager
-from typing import Callable, Collection, Generator, TypeVar
-
-T = TypeVar("T")
+from typing import Callable, Collection, Generator, Any
 
 
 @contextmanager
@@ -19,8 +17,8 @@ def wrap_thread_pool(pool: ThreadPoolExecutor) -> Generator[ThreadPoolExecutor, 
 
 
 def call_in_parallel(
-    callables: Collection[Callable[[], T]], timeout: float, max_threads: int = 10
-) -> Generator[Future[T], None, None]:
+    callables: Collection[Callable[[], Any]], timeout: float, max_threads: int = 10
+) -> Generator[Future, None, None]:
     """
     Call the given callables in parallel and generate their futures in correspondence to their finishing order.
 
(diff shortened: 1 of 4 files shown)


granulate_utils/metadata/cloud.py
```diff
@@ -167,9 +168,9 @@ def get_static_cloud_instance_metadata(logger: Union[logging.LoggerAdapter, logg
     def _fetch() -> Tuple[Optional[Metadata], List[Exception]]:
         cloud_metadata_fetchers = [get_aws_metadata, get_gcp_metadata, get_azure_metadata]
         raised_exceptions: List[Exception] = []
-        for fetcher in cloud_metadata_fetchers:
+        for future in call_in_parallel(cloud_metadata_fetchers, timeout=6):
             try:
-                response = fetcher()
+                response = future.result()
```
michelhe, line 173: Aren't the other futures leaked now if the first completed one was successful?

netaneld122: They are leaked until they finish running in the background, which is why I documented that it should be ok as long as the callables are limited in time.

In terms of object/memory leak, I don't think that's the case:
```
In [13]: len(gc.get_objects())
Out[13]: 142881

In [14]: next(call_in_parallel([lambda: 1 for _ in range(1000000)], timeout=10)).result()
Out[14]: 1

In [15]: len(gc.get_objects())
Out[15]: 141374

In [16]: next(call_in_parallel([lambda: i for i in range(1000000)], timeout=10)).result()
Out[16]: 999999

In [17]: len(gc.get_objects())
Out[17]: 141410
```

Jongy (commented):
Cool :) I'll review this week

granulate_utils/futures.py
```diff
@@ -0,0 +1,32 @@
...
+
+T = TypeVar("T")
+
+
+@contextmanager
+def wrap_thread_pool(pool: ThreadPoolExecutor) -> Generator[ThreadPoolExecutor, None, None]:
+    try:
+        yield pool
+    except _base.TimeoutError as e:
+        raise TimeoutError(*e.args) from e
```
Jongy, line 13: What's this?

netaneld122: For some reason they decided to implement their own version of TimeoutError, and only fixed it in python 3.11, so I translate their TimeoutError to the standard one:
image
I'll document it.

granulate_utils/futures.py
```diff
@@ -0,0 +1,38 @@
...
+        # so we achieve here the same behavior for older python versions
+        # https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.TimeoutError
+        raise TimeoutError(*e.args) from e
+    finally:
+        pool.shutdown(wait=False)
+
+
+def call_in_parallel(
+    callables: Collection[Callable[[], T]], timeout: float, max_threads: int = 10
+) -> Generator[Future[T], None, None]:
```
Jongy, line 23: I tried to run it with gProfiler - 
```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/app/gprofiler/__main__.py", line 5, in <module>
    from gprofiler.main import main
  File "/app/gprofiler/main.py", line 46, in <module>
    from gprofiler.metadata.metadata_collector import get_current_metadata, get_static_metadata
  File "/app/gprofiler/metadata/metadata_collector.py", line 4, in <module>
    from granulate_utils.metadata.cloud import get_static_cloud_instance_metadata
  File "/usr/local/lib/python3.8/dist-packages/granulate_utils/metadata/cloud.py", line 16, in <module>
    from granulate_utils.futures import call_in_parallel
  File "/usr/local/lib/python3.8/dist-packages/granulate_utils/futures.py", line 23, in <module>
    ) -> Generator[Future[T], None, None]:
TypeError: 'type' object is not subscriptable
```

it's 3.8 (gProfiler container)

netaneld122: Ahh shit, we should run the tests on all python versions 😢

tests/granulate_utils/test_futures.py
```diff
@@ -0,0 +1,36 @@
...
+    for i, future in enumerate(call_in_parallel((lambda: wait(0), lambda: wait(1), lambda: wait(2)), timeout=5)):
+        assert i == future.result()
+
+
+def test_call_in_parallel_timeout() -> None:
+    with pytest.raises(TimeoutError, match="futures unfinished"):
+        next(call_in_parallel((lambda: wait(2), lambda: wait(2)), timeout=1)).result()
+
+
+def test_call_in_parallel_exception_handling() -> None:
```
Jongy, line 25: One more test I can think of (and actually wonder what the result would be) is if we have one that throws and then we call next again.
This happens in your change and actually I wonder if it's correct, you built it as 
```
for x in call_in_parallel:
   try:
     x.result()
   except ....
```

To my understanding the `next` is called in the `for` line, so outside of the `try`. It's the `result` call that's raising I suppose, but let's test it

granulate_utils/futures.py
Jongy: We do run linting on python 3.8 but apparently this isn't caught

granulate_utils/metadata/cloud.py
Jongy: It's okay after the TPE and threads die

granulate_utils/futures.py
netaneld122: Using version matrix for testing as well now so we don't accidentally introduce bugs missed by the linter

Jongy (dismissed):
Lgtm. I'll test tmrw with the profiler, then we merge

Jongy:
@netaneld122 open PR in gprofiler to update?

PR #134, opened 2023-03-22
is_musl(): Accept memory maps as an optional param and explain about glibc-compat

Jongy:


d3dave (commented):
Why the maps argument?

Jongy:
In https://github.com/Granulate/gprofiler/pull/737 I use it to cache `memory_maps` call, because gProfiler now checks for `musl` and also for `glibc-compat`. Wanted to avoid 2 calls.

PR #135, opened 2023-03-23
gProfiler's spark collector migration

roi-granulate:
Copied from https://github.com/Granulate/gprofiler/tree/69d98b917220015b65cf481bde8536a81e4be92b

### DIFF
--- a/granulate_utils/metrics/collector.py
+++ b/granulate_utils/metrics/collector.py
@@ -1,0 +1,484 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+# (C) Datadog, Inc. 2018-present. All rights reserved.
+# Licensed under a 3-clause BSD style license (see LICENSE.bsd3).
+#
+import re
+from typing import Any, Dict, Iterable, List, Tuple, Union
+from urllib.parse import urljoin, urlparse
+
+import requests
+from bs4 import BeautifulSoup
+from requests import HTTPError
+
+# Application type and states to collect
+from granulate_utils.metrics import Sample
+from granulate_utils.metrics.metrics import (
+    SPARK_AGGREGATED_STAGE_METRICS,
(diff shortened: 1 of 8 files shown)


granulate_utils/metrics/sampler.py
```diff
@@ -0,0 +1,240 @@
...
+            assert os.path.exists(self._storage_dir) and os.path.isdir(self._storage_dir)
+        else:
+            self._logger.debug("Output directory is None. Will add metrics to queue")
+
+    def get_hostname(self) -> str:
+        try:
+            hostname = socket.gethostname()
+        except Exception as e:
+            self._logger.exception("Failed to get hostname", exception=e)
+        return hostname
```
Jongy, line 52: This is incorrect to get hostname that way. gProfiler gets the hostname in the outer network & mount namespace.

Let's keep this code outside of this class - it should accept hostname as a parameter, then gProfiler can pass the hostname it has.

Generally when moving code - DO NOT resort to re-implementing things because it's inconvenient to move them. Let's discuss each particular case.

granulate_utils/metrics/sampler.py
```diff
@@ -0,0 +1,240 @@
...
+class SparkSampler:
+    """
+    Spark cluster metrics sampler
+    """
+
+    def __init__(
+        self,
+        sample_period: float,
+        storage_dir: str,
+        logger: Any,
```
Jongy, line 30: Use a proper type.

granulate_utils/metrics/collector.py
```diff
@@ -0,0 +1,484 @@
...
+
+STRUCTURED_STREAMS_METRICS_REGEX = re.compile(
+    r"^[\w-]+\.driver\.spark\.streaming\.(?P<query_name>[\w-]+)\.(?P<metric_name>[\w-]+)$"
+)
+
+
+class SparkCollector:
+    def __init__(
+        self,
+        logger: Any,
```
Jongy, line 53: Non Any logger please

Jongy (changes requested):
Done reviewing.

PR #138, opened 2023-04-09
Big data collectors fix

roi-granulate:


granulate_utils/metrics/sampler.py
```diff
@@ -65,21 +65,13 @@ def __init__(
         self._hostname = hostname
         self._applications_metrics = applications_metrics
         self._spark_samplers: List[Collector] = []
+        self._master_address: Optional[str] = None
+        self._cluster_mode: Optional[str] = None
 
         if (cluster_mode is not None) and (master_address is not None):
```
Jongy, line 71: If both must be given, it's an error to give only one. `assert` here? Otherwise, your function accepts arguments that it silently ignores.

granulate_utils/metrics/sampler.py
```diff
@@ -283,24 +280,28 @@ def discover(self) -> bool:
...
+        if self._master_address is not None and self._cluster_mode is not None:
+            # No need to guess, manually configured
+            self._logger.debug(
+                "No need to guess cluster mode and master address, manually configured",
+                cluster_mode=self._cluster_mode,
+                master_address=self._master_address,
+            )
+            have_conf = True
+
+        elif self._master_address is None or self._cluster_mode is None:
```
Jongy, line 293:
```suggestion
        else:
```
?

granulate_utils/metrics/sampler.py
```diff
@@ -269,7 +261,12 @@ def _init_collectors(self):
         """
         if self._cluster_mode == SPARK_YARN_MODE:
             self._spark_samplers.append(YarnCollector(self._master_address, self._logger))
-        elif self._cluster_mode == SPARK_STANDALONE_MODE or self._cluster_mode == SPARK_MESOS_MODE:
+
+        # In Standalone and Mesos we'd use applications metrics
+        if self._cluster_mode in (SPARK_STANDALONE_MODE, SPARK_MESOS_MODE):
+            self._applications_metrics = True
+
+        if self._applications_metrics is True:
```
Jongy, line 269: Do not use `is True` / `is False`. It is not customary.

Jongy (changes requested):

granulate_utils/metrics/sampler.py
roi-granulate: 493f35963d9fdb4056e587f57176efbebc31606c

granulate_utils/metrics/sampler.py
roi-granulate: 3e468172567a6ac993ce64602a0f9580aae7cb40

granulate_utils/metrics/sampler.py
roi-granulate: e43731f86d25e2832dfba862812b2d96ba8db67a

Jongy:
Looks good now, we'll merge after gprofiler side is ready as well.

Jongy (approved):

Jongy:
Let's merge it

PR #139, opened 2023-04-15
Upgrade packages, take 2

Jongy:


requirements.txt
```diff
@@ -1,9 +1,8 @@
 psutil~=5.8.0
 requests~=2.27.1
-grpcio~=1.43.0
+grpcio~=1.53.0
 protobuf~=3.19.4
 docker~=5.0.0
-dataclasses~=0.8; python_version < '3.7'
```
Jongy, line 6: We no longer support 3.7 so this is not required.

requirements.txt
```diff
@@ -1,9 +1,8 @@
 psutil~=5.8.0
 requests~=2.27.1
-grpcio~=1.43.0
+grpcio~=1.53.0
```
Jongy, line 3: The main motive for it is upgrading gprofiler to python 3.11 which requires newer grpcio.

Jongy:
Pending test - I will run gProfiler on a box which runs k8s CRI and ensure gProfiler is able to use the `ContainersClient` (CRI) with the new grpc version.

Jongy:
I am waiting for next grcpio releases to ensure it doesn't hang gProfiler anymore.

PR #143, opened 2023-04-19
Bigdata better discover API

roi-granulate:
https://github.com/Granulate/gprofiler/pull/749#discussion_r1169921653

### DIFF
--- a/granulate_utils/metrics/sampler.py
+++ b/granulate_utils/metrics/sampler.py
@@ -283,6 +283,8 @@ class BigDataSampler(Sampler):
 
         returns True if we have these configurations, False otherwise
         """
+        assert self._spark_samplers == [], "discover() should only be called once"
+
         have_conf = False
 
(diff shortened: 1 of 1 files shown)


granulate_utils/metrics/sampler.py
```diff
@@ -307,7 +309,7 @@ def discover(self) -> bool:
                 )
                 have_conf = True
 
-        if have_conf and self._spark_samplers == []:
+        if have_conf and self._collectors == []:
```
Jongy, line 312: This makes no sense - you added an `assert` for it in line 286. What you probably need to assert for is that `_init_collectors` was not called, so you can keep the `collectors == []` assert, and remove the check here.

Jongy (changes requested):
.

granulate_utils/metrics/sampler.py
roi-granulate: forgot to commit that

Jongy (approved):

PR #144, opened 2023-04-27
Update grpcio to 1.53

mpozniak95:


granulate_utils/containers/cri.py
```diff
@@ -38,9 +39,10 @@ class CriClient(ContainersClientInterface):
     def __init__(self) -> None:
         self._runtimes = {}
         for rt, path in RUNTIMES:
-            path = "unix://" + ns.resolve_host_root_links(path)
-            if self._is_cri_available(path):
-                self._runtimes[rt] = path
+            if os.path.exists(path):
```
marcin-ol, line 42: Not sure what better we can do about it.
It seems reasonable to put this check here, and prevent grpc from retrying connection to unavailable unix socket.

marcin-ol:
Did some research on grpcio, tried versions as recent as 1.56.2.  

Without the condition added by Marcin, grpcio seems to retry connection to unavailable unix socket and occupy grpc thread.

Here's a fragment of GRPC trace:
```
I0809 10:20:47.309468319 1277687 subchannel.cc:707]                    subchannel 0x55d9d6b7acb0 {address=unix:/proc/1/root/run/crio/crio.sock, 
grpc.primary_user_agent=grpc-python/1.56.2, grpc.resource_quota=0x55d9d6a85910, grpc.server_uri=unix:///proc/1/root/run/crio/crio.sock}}:
connect failed (UNKNOWN:No such file or directory {created_time:"2023-08-09T10:20:47.309356824+00:00",   
errno:2, os_error:"No such file or directory", syscall:"connect", target_address:"unix:/proc/1/root/run/crio/crio.sock"}),   
backing off for 1000 ms
```

The issue will only surface if we actually attempt to send a grpc Version request in method `_is_cri_available()`.

Playing with grpc options found the following to help with some cases, but not eliminate the problem of hanging-processes completely:
```python
self._channel = grpc.insecure_channel(path, options=[
    ('grpc.initial_reconnect_backoff_ms', 100,),
    ('grpc.min_reconnect_backoff_ms', 100,),
    ('grpc.max_reconnect_backoff_ms', 100,),
])

```

Jongy:
Closing in favor of https://github.com/Granulate/granulate-utils/pull/139

PR #148, opened 2023-05-04
Copy gProfilers' Databricks client with minor adjustments

roi-granulate:
for review convenience
exact copy from gprofiler f120423
minor adjustment of logger 746a5e6


Jongy (approved):
LGTM, thanks @roi-granulate for the clean move, was easy to review

PR #153, opened 2023-06-04
add auth method params

DaveAvigad:
this PR adds authentication parameters to the Sender requests to allow G-Logger Backend to authenticate the request using aws token **or** g-profiler token. 

new params:
-  x_auth_type: The type of authentication to use. ('aws')
-  x_auth_access_key_id: The access key id to use for authentication.
-  x_auth_secret_access_key: The secret access key to use for authentication.

### DIFF
--- a/glogger/sender.py
+++ b/glogger/sender.py
@@ -38,7 +38,7 @@ class Sender:
         server_address: str,
         *,
         scheme: str = "https",
-        auth_token_type: str = "",
+        auth_secret: str = "",
         send_interval: float = 30.0,
         send_threshold: float = 0.8,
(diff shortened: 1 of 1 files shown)


michelhe (dismissed):
LGTM

glogger/sender.py
```diff
@@ -38,6 +38,9 @@ def __init__(
         server_address: str,
         *,
         scheme: str = "https",
+        x_auth_type: str = "",
```
DaveAvigad, line 41: remove the commas else it will change the type into a tupple

michelhe: You must be confused.

PR #154, opened 2023-06-05
upgrade grpcio

odelame:


d3dave (approved):

Jongy:
See https://github.com/Granulate/granulate-utils/pull/139 and https://github.com/Granulate/granulate-utils/pull/144 and the problems it has caused in https://github.com/Granulate/gprofiler/pull/775#issuecomment-1515996235. Do not merge it

PR #156, opened 2023-06-12
DBX wrapper refactor

roi-granulate:
note there are some comments I left to ease the review process

### DIFF
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -25,9 +25,7 @@ CLUSTER_NAME_KEY = "ClusterName"
 DEFAULT_WEBUI_PORT = 40001
 DATABRICKS_JOBNAME_TIMEOUT_S = 2 * 60
 RETRY_INTERVAL_S = 1
-
 RUN_ID_REGEX = "run-\\d+-"
-
 
 class DBXWebUIEnvWrapper:
     def __init__(self, logger: logging.LoggerAdapter) -> None:
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -69,7 +69,7 @@ class DBXWebUIEnvWrapper:
                     )
                     return cluster_all_props
                 else:
-                    # No job name yet, retry.
+                    # No environment metadata yet, retry.
                     time.sleep(RETRY_INTERVAL_S)
             except DatabricksJobNameDiscoverException:
                 self.logger.exception("Failed to get Databricks job name")
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -173,7 +173,7 @@ class DBXWebUIEnvWrapper:
         return metadata
 
     @property
-    def get_name_from_metadata(self) -> Optional[str]:
+    def name(self) -> Optional[str]:
         assert self.all_props_dict is not None, "all_props_dict is None, can't get name from metadata"
         if job_name := self.all_props_dict.get(JOB_NAME_KEY):
             return f"job-{job_name}"
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -174,7 +174,7 @@ class DBXWebUIEnvWrapper:
 
         # First, trying to extract `CLUSTER_TAGS_KEY` property, in case not redacted.
         if (
-                cluster_all_tags_value := service_name_prop_candidates.get(CLUSTER_ALL_TAGS_PROP)
+            cluster_all_tags_value := service_name_prop_candidates.get(CLUSTER_ALL_TAGS_PROP)
         ) is not None and "redacted" not in cluster_all_tags_value:
             try:
                 cluster_all_tags_value_json = json.loads(cluster_all_tags_value)
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -172,10 +172,10 @@ class DBXWebUIEnvWrapper:
             metadata[CLUSTER_NAME_KEY] = metadata[CLUSTER_NAME_KEY].replace(" ", "-").lower()
         return metadata
 
-
-def get_name_from_metadata(metadata: Dict[str, str]) -> Optional[str]:
-    if JOB_NAME_KEY in metadata:
-        return f"job-{metadata[JOB_NAME_KEY]}"
-    elif CLUSTER_NAME_KEY in metadata:
-        return metadata[CLUSTER_NAME_KEY]
-    return None
+    @property
+    def get_name_from_metadata(self) -> Optional[str]:
+        if job_name := self.all_props_dict.get(JOB_NAME_KEY):
+            return f"job-{job_name}"
+        elif cluster_name := self.all_props_dict.get(CLUSTER_NAME_KEY):
+            return cluster_name
+        return None
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -144,19 +144,19 @@ class DBXWebUIEnvWrapper:
                 cluster_all_tags_value_json = json.loads(cluster_all_tags_value)
             except Exception as e:
                 raise DatabricksJobNameDiscoverException(f"Failed to parse {cluster_all_tags_value}") from e
-            return self._enforce_pattern(
+            return self._apply_pattern(
                 {cluster_all_tag["key"]: cluster_all_tag["value"] for cluster_all_tag in cluster_all_tags_value_json}
             )
         # As a fallback, trying to extract `CLUSTER_NAME_PROP` property.
         elif (cluster_name_value := service_name_prop_candidates.get(CLUSTER_NAME_PROP)) is not None:
-            return self._enforce_pattern({CLUSTER_NAME_KEY: cluster_name_value})
+            return self._apply_pattern({CLUSTER_NAME_KEY: cluster_name_value})
         else:
             raise DatabricksJobNameDiscoverException(
                 f"Failed to extract {CLUSTER_ALL_TAGS_PROP} or {CLUSTER_NAME_PROP} from {props=}"
             )
 
     @staticmethod
-    def _enforce_pattern(metadata: Dict[str, str]) -> Dict[str, str]:
+    def _apply_pattern(metadata: Dict[str, str]) -> Dict[str, str]:
         """
         This function is used to enforce certain regex and other patterns on some metadata values, on which we
         know problematic to include in service names.
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -84,7 +84,7 @@ class DBXWebUIEnvWrapper:
                 return None
             if not self.enable_retries:
                 break
-        self.logger.info("Databricks get job name timeout, continuing...")
+        self.logger.info("Databricks get DBX environment metadata timeout, continuing...")
         return None
 
     def _discover_apps_url(self):
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -84,6 +84,7 @@ class DBXWebUIEnvWrapper:
     def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
         """
         Returns `includes spark.databricks.clusterUsageTags.clusterAllTags` tags as `Dict`.
+        In any case this function returns `None`, a retry is required.
         """
         if not os.path.isfile(DATABRICKS_METRICS_PROP_PATH):
             # We want to retry in case the cluster is still initializing, and the file is not yet deployed.
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -130,8 +130,9 @@ class DBXWebUIEnvWrapper:
         if props is None:
             raise DatabricksJobNameDiscoverException(f"sparkProperties was not found in {env=}")
         # Creating a dict of the relevant properties and their values.
+        relevant_props = [CLUSTER_ALL_TAGS_PROP, CLUSTER_NAME_PROP]
         service_name_prop_candidates = {
-            prop[0]: prop[1] for prop in props if (CLUSTER_ALL_TAGS_PROP == prop[0] or CLUSTER_NAME_PROP == prop[0])
+            prop[0]: prop[1] for prop in props if prop[0] in relevant_props
         }
         if len(service_name_prop_candidates) == 0:
             # We expect at least one of the properties to be present.
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -6,6 +6,7 @@
 import json
 import logging
 import os
+import re
 import time
 from typing import Dict, Optional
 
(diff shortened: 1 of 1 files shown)


granulate_utils/metadata/databricks_client.py
```diff
@@ -135,12 +128,49 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
         props = env.get("sparkProperties")
         if props is None:
             raise DatabricksJobNameDiscoverException(f"sparkProperties was not found in {env=}")
-        for prop in props:
-            if prop[0] == CLUSTER_TAGS_KEY:
-                try:
-                    all_tags_value = json.loads(prop[1])
-                except Exception as e:
-                    raise DatabricksJobNameDiscoverException(f"Failed to parse {prop=}") from e
-                return {cluster_all_tag["key"]: cluster_all_tag["value"] for cluster_all_tag in all_tags_value}
```
roi-granulate, line 144: This dict is now contained in `relevant_props_dict`

granulate_utils/metadata/databricks_client.py
```diff
@@ -56,21 +59,18 @@ def get_webui_address() -> Optional[str]:
...
+                    self.logger.info(
+                        "Successfully got relevant cluster tags metadata", cluster_all_props=cluster_all_props
+                    )
+                    return cluster_all_props
                 else:
-                    # No job name yet, retry.
+                    # No environment metadata yet, retry.
                     time.sleep(RETRY_INTERVAL_S)
             except DatabricksJobNameDiscoverException:
                 self.logger.exception("Failed to get Databricks job name")
```
mirikl, line 76: These log messages should also be changed right?

mirikl (changes requested):
Not done yet

granulate_utils/metadata/databricks_client.py
roi-granulate: 67bb152

granulate_utils/metadata/databricks_client.py
```diff
@@ -135,12 +129,50 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
+        elif (cluster_name_value := service_name_prop_candidates.get(CLUSTER_NAME_PROP)) is not None:
+            return self._enforce_pattern({CLUSTER_NAME_KEY: cluster_name_value})
         else:
-            raise DatabricksJobNameDiscoverException(f"Failed to find {CLUSTER_TAGS_KEY=} in {props=}")
+            raise DatabricksJobNameDiscoverException(
+                f"Failed to extract {CLUSTER_ALL_TAGS_PROP} or {CLUSTER_NAME_PROP} from {props=}"
+            )
+
+    @staticmethod
+    def _enforce_pattern(metadata: Dict[str, str]) -> Dict[str, str]:
```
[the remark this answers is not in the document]
roi-granulate: 7ce1b51

granulate_utils/metadata/databricks_client.py
```diff
@@ -135,12 +129,50 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
+            metadata[JOB_NAME_KEY] = metadata[JOB_NAME_KEY].replace(" ", "-").lower()
+        if CLUSTER_NAME_KEY in metadata:
+            # We've tackled cases where the cluster name includes Run ID, we want to remove it.
+            metadata[CLUSTER_NAME_KEY] = re.sub(RUN_ID_REGEX, "", metadata[CLUSTER_NAME_KEY])
+            metadata[CLUSTER_NAME_KEY] = metadata[CLUSTER_NAME_KEY].replace(" ", "-").lower()
+        return metadata
+
+
+def get_name_from_metadata(metadata: Dict[str, str]) -> Optional[str]:
+    if JOB_NAME_KEY in metadata:
```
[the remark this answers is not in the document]
roi-granulate: b7ccd66

granulate_utils/metadata/databricks_client.py
```diff
@@ -135,12 +129,50 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
+        if JOB_NAME_KEY in metadata:
+            metadata[JOB_NAME_KEY] = metadata[JOB_NAME_KEY].replace(" ", "-").lower()
+        if CLUSTER_NAME_KEY in metadata:
+            # We've tackled cases where the cluster name includes Run ID, we want to remove it.
+            metadata[CLUSTER_NAME_KEY] = re.sub(RUN_ID_REGEX, "", metadata[CLUSTER_NAME_KEY])
+            metadata[CLUSTER_NAME_KEY] = metadata[CLUSTER_NAME_KEY].replace(" ", "-").lower()
+        return metadata
+
+
+def get_name_from_metadata(metadata: Dict[str, str]) -> Optional[str]:
```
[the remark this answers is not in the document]
roi-granulate: b7ccd66

granulate_utils/metadata/databricks_client.py
```diff
@@ -135,12 +129,50 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
+                f"Failed to extract {CLUSTER_ALL_TAGS_PROP} or {CLUSTER_NAME_PROP} from {props=}"
+            )
+
+    @staticmethod
+    def _enforce_pattern(metadata: Dict[str, str]) -> Dict[str, str]:
+        """
+        This function is used to enforce certain regex and other patterns on some metadata values, on which we
+        know problematic to include in service names.
+        """
+        if JOB_NAME_KEY in metadata:
```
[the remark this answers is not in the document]
roi-granulate: I don't think it is necessary at this point.
but in case we want to make things more complex:
we can divide the service pattern apply into two families, regex (with re.sub) and `str` manipulations (`.replace()` and `lower()` etc.)
as with string manipulation we'd like to apply on all values I guess.
as with regex, we might want to have a dict of regex compilations to sub, where:
`{<prop key>: <regex>}` 
e.g.: `apply_regex_by_prop_dict = {CLUSTER_NAME_KEY: RUN_ID_REGEX}`

granulate_utils/metadata/databricks_client.py
```diff
@@ -135,12 +129,54 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
+            metadata[JOB_NAME_KEY] = metadata[JOB_NAME_KEY].replace(" ", "-").lower()
+        if CLUSTER_NAME_KEY in metadata:
+            # We've tackled cases where the cluster name includes Run ID, we want to remove it.
+            metadata[CLUSTER_NAME_KEY] = re.sub(RUN_ID_REGEX, "", metadata[CLUSTER_NAME_KEY])
+            metadata[CLUSTER_NAME_KEY] = metadata[CLUSTER_NAME_KEY].replace(" ", "-").lower()
+        return metadata
+
+    @property
+    def get_name_from_metadata(self) -> Optional[str]:
```
mirikl, line 176: I prefer if the property itself won't have a `get_` inside it. Maybe something like this:
```suggestion
    def job_name(self) -> Optional[str]:
```

granulate_utils/metadata/databricks_client.py
mirikl: There's another log line under it also :)

granulate_utils/metadata/databricks_client.py
```diff
@@ -135,12 +129,54 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
         if props is None:
             raise DatabricksJobNameDiscoverException(f"sparkProperties was not found in {env=}")
-        for prop in props:
-            if prop[0] == CLUSTER_TAGS_KEY:
-                try:
-                    all_tags_value = json.loads(prop[1])
-                except Exception as e:
-                    raise DatabricksJobNameDiscoverException(f"Failed to parse {prop=}") from e
-                return {cluster_all_tag["key"]: cluster_all_tag["value"] for cluster_all_tag in all_tags_value}
+        # Creating a dict of the relevant properties and their values.
```
mirikl, line 132:
```suggestion
        # Creating a dict of the relevant properties and their values.
        relevant_props = [CLUSTER_ALL_TAGS_PROP, CLUSTER_NAME_PROP]
        service_name_prop_candidates = {
            prop[0]: prop[1] for prop in props if prop[0] in relevant_props
        }
```

granulate_utils/metadata/databricks_client.py
```diff
@@ -135,12 +129,54 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
+            try:
+                cluster_all_tags_value_json = json.loads(cluster_all_tags_value)
+            except Exception as e:
+                raise DatabricksJobNameDiscoverException(f"Failed to parse {cluster_all_tags_value}") from e
+            return self._apply_pattern(
+                {cluster_all_tag["key"]: cluster_all_tag["value"] for cluster_all_tag in cluster_all_tags_value_json}
+            )
+        # As a fallback, trying to extract `CLUSTER_NAME_PROP` property.
+        elif (cluster_name_value := service_name_prop_candidates.get(CLUSTER_NAME_PROP)) is not None:
+            return self._apply_pattern({CLUSTER_NAME_KEY: cluster_name_value})
```
mirikl, line 152: Is this a key that might be present in `cluster_all_tags_value_json` ? If not, then what about having a different `_apply_pattern` for `JOB_NAME_KEY` and another for `CLUSTER_NAME_KEY`?

granulate_utils/metadata/databricks_client.py
```diff
@@ -135,12 +129,54 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
+        ) is not None and "redacted" not in cluster_all_tags_value:
+            try:
+                cluster_all_tags_value_json = json.loads(cluster_all_tags_value)
+            except Exception as e:
+                raise DatabricksJobNameDiscoverException(f"Failed to parse {cluster_all_tags_value}") from e
+            return self._apply_pattern(
+                {cluster_all_tag["key"]: cluster_all_tag["value"] for cluster_all_tag in cluster_all_tags_value_json}
+            )
+        # As a fallback, trying to extract `CLUSTER_NAME_PROP` property.
+        elif (cluster_name_value := service_name_prop_candidates.get(CLUSTER_NAME_PROP)) is not None:
```
mirikl, line 151:
```suggestion
        elif cluster_name_value := service_name_prop_candidates.get(CLUSTER_NAME_PROP):
```
?

mirikl (changes requested):
Are there any UTs we can write for this code?

granulate_utils/metadata/databricks_client.py
roi-granulate: 3f8726c

granulate_utils/metadata/databricks_client.py
roi-granulate: 2a71205

granulate_utils/metadata/databricks_client.py
roi-granulate: yes, `CLUSTER_NAME_KEY` can appear in both

granulate_utils/metadata/databricks_client.py
roi-granulate: DM'ed

granulate_utils/metadata/databricks_client.py
```diff
@@ -132,15 +130,89 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
+        """
+        Returns `includes spark.databricks.clusterUsageTags.clusterAllTags` tags as `Dict`.
+        In any case this function returns `None`, a retry is required.
+        """
+        if not os.path.isfile(DATABRICKS_METRICS_PROP_PATH):
+            # We want to retry in case the cluster is still initializing, and the file is not yet deployed.
+            return None
+        # Discovering SparkUI apps url.
+        self._discover_apps_url()
+        if self._is_apps_url_discovered() is False:
```
mirikl, line 157:
```suggestion
        if not self._is_apps_url_discovered():
```

granulate_utils/metadata/databricks_client.py
```diff
@@ -56,53 +59,49 @@ def get_webui_address() -> Optional[str]:
...
+        if self._is_apps_url_discovered() is True:
+            return True
+        else:
+            if (web_ui_address := self.get_webui_address()) is None:
+                return False
+            self._apps_url = SPARKUI_APPS_URL.format(web_ui_address)
+            self.logger.debug("Databricks SparkUI address", apps_url=self._apps_url)
+            return True
+
+    def _is_apps_url_discovered(self):
```
mirikl, line 98: I think this function is redundant

granulate_utils/metadata/databricks_client.py
```diff
@@ -56,53 +59,49 @@ def get_webui_address() -> Optional[str]:
...
         return None
 
-    @staticmethod
-    def _get_name_from_metadata(metadata: Dict[str, str]) -> Optional[str]:
-        if JOB_NAME_KEY in metadata:
-            return str(metadata[JOB_NAME_KEY]).replace(" ", "-").lower()
-        return None
-
-    def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
+    def _discover_apps_url(self) -> bool:
```
mirikl, line 84: What about changing this function to be `_get_apps_url` and it will return an `Optional[str]` ? Then we can just call it from `_cluster_all_tags_metadata` and check if we were able to get it?

granulate_utils/metadata/databricks_client.py
mirikl: Or if you take my suggestion from the previous comments this can just be:
```
    if (apps_url := self._get_apps_url()) is None:
        # SparkUI apps url was not discovered, retrying.
        return None
```

granulate_utils/metadata/databricks_client.py
```diff
@@ -132,15 +130,89 @@ def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
...
+            return None
+
+        # Getting spark apps in JSON format.
+        apps = self._spark_apps_json()
+        if len(apps) == 0:
+            # apps might be empty because of initialization, retrying.
+            self.logger.debug("No apps yet, retrying.")
+            return None
+
+        # Extracting for the first app the "sparkProperties" table of the application environment.
```
mirikl, line 168: Consider extracting this section into a sub-function as well, this is really nit-picking feel free to ignore.

mirikl (changes requested):
Small stuff

granulate_utils/metadata/databricks_client.py
roi-granulate: 2ed7a37

granulate_utils/metadata/databricks_client.py
roi-granulate: 8f44920

granulate_utils/metadata/databricks_client.py
roi-granulate: We're discovering an object property
we should discover it, and validate it was discovered.
I don't think we should call `_get_apps_url` where we do know that the property might not have a value yet.
please LMK if you still think differently

granulate_utils/metadata/databricks_client.py
```diff
@@ -56,53 +63,48 @@ def get_webui_address() -> Optional[str]:
...
         return None
 
-    @staticmethod
-    def _get_name_from_metadata(metadata: Dict[str, str]) -> Optional[str]:
-        if JOB_NAME_KEY in metadata:
-            return str(metadata[JOB_NAME_KEY]).replace(" ", "-").lower()
-        return None
-
-    def _cluster_all_tags_metadata(self) -> Optional[Dict[str, str]]:
+    def _discover_apps_url(self):
```
mirikl, line 90: The boolean this function returned is never checked, so maybe remove it? If you keep it then please add type hinting.

mirikl (changes requested):
1 small

granulate_utils/metadata/databricks_client.py
roi-granulate: 0d528f0

mirikl (approved):

PR #160, opened 2023-06-23
Add method to check if exe is stripped

mpozniak95:
Connected with: https://github.com/Granulate/gprofiler/issues/804.

As far as I understand if executable has debug information it is not-stripped, and if it hasn't then it is stripped. According to pyelftools this information is stored in has_debug_info property in DWARFInfo object which can be obtained by using get_dwarf_info() method on elf file.

### DIFF
--- a/granulate_utils/linux/elf.py
+++ b/granulate_utils/linux/elf.py
@@ -183,3 +183,10 @@ def get_libc_type(elf: ELFType) -> LibcType:
             return LibcType.STATIC_LIBC
 
         return LibcType.STATIC_NO_LIBC
+
+
+def elf_has_debug_info(elf: ELFType) -> bool:
+    with open_elf(elf) as elf:
+        if elf.has_dwarf_info():
+            return elf.get_dwarf_info().has_debug_info
+    return False
--- a/granulate_utils/linux/elf.py
+++ b/granulate_utils/linux/elf.py
@@ -185,7 +185,7 @@ def get_libc_type(elf: ELFType) -> LibcType:
         return LibcType.STATIC_NO_LIBC
 
 
-def elf_is_stripped(elf: ELFType) -> None:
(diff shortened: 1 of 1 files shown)


mpozniak95:
@Jongy Please take a look

Jongy (commented):
I am not sure that this is the correct property. I suggest trying to compile a dummy program with gcc, run "file" and see that it gives "not stripped". Then run "strip" on it and now you will get "stripped". Your function should behave accordingly

mpozniak95:
Ok, you are right then. This is not correct property. I'll search for something else

granulate_utils/linux/elf.py
```diff
@@ -183,3 +183,10 @@ def get_libc_type(elf: ELFType) -> LibcType:
             return LibcType.STATIC_LIBC
 
         return LibcType.STATIC_NO_LIBC
+
+
+def elf_is_stripped(elf: ELFType) -> bool:
```
Jongy, line 188: This makes sense, I assume you tested it per my suggestions [here](https://github.com/Granulate/granulate-utils/pull/160#pullrequestreview-1494813829)?

granulate_utils/linux/elf.py
```diff
@@ -183,3 +183,10 @@ def get_libc_type(elf: ELFType) -> LibcType:
...
+    with open_elf(elf) as elf:
+        if elf.get_section_by_name(".symtab") is not None:
+            return False
+        return True
```
Jongy, line 192:
```suggestion
        return elf.get_section_by_name(".symtab") is None
```
saner

Jongy (changes requested):
.

granulate_utils/linux/elf.py
mpozniak95: Yes I tested it

Jongy (approved):

PR #166, opened 2023-07-09
Add cluster_id to Spark/YARN collector

rgruner:
1. Add `cluster_id` label to identify different clusters in the spark/yarn collector. 
2. We're using the Hadoop Yarn cluster identifier to uniquely identify a cluster. This means that mostly Yarn is supported, preliminary support for other cluster types is implemented (using the master host name)

### DIFF
--- a/granulate_utils/metrics/sampler.py
+++ b/granulate_utils/metrics/sampler.py
@@ -266,25 +266,22 @@ class BigDataSampler(Sampler):
         This function fills in self._spark_samplers with the appropriate collectors.
         """
         if self._cluster_mode == SPARK_YARN_MODE:
-            self._collectors.append(YarnCollector(self._master_address, self._logger))
+            yarn_collector = YarnCollector(self._master_address, self._logger)
+            self._collectors.append(yarn_collector)
+            self._cluster_id = self._cluster_id or yarn_collector.cluster_id
 
-        # In Standalone and Mesos we'd use applications metrics
-        if self._cluster_mode in (SPARK_STANDALONE_MODE, SPARK_MESOS_MODE):
+        elif self._cluster_mode in (SPARK_STANDALONE_MODE, SPARK_MESOS_MODE):
+            # In Standalone and Mesos we'd use applications metrics
             self._applications_metrics = True
+
+            # Didn't research how to bring the actual cluster id, so use the master address
+            # Which should work for some use cases.
+            self._cluster_id = self._cluster_id or self._master_address
 
         if self._applications_metrics:
             self._collectors.append(
                 SparkApplicationMetricsCollector(self._cluster_mode, self._master_address, self._logger)
             )
-
-    def _get_cluster_id(self) -> Optional[str]:
-        if self._cluster_mode == SPARK_YARN_MODE and self._master_address is not None:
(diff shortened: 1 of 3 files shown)


roi-granulate:
I personally don't like the idea that `YarnCollector` instance is being transferred to `SparkApplicationMetricsCollector`
I don't think there is a reason to transfer collectors between themselves as arguments
You can enforce `Sampler` to have `self._id` - or just have it in `BigDataSampler` (no `Sampler` enforcement), which will practically behave as a cluster ID, and then call a function from `Sampler.collect()` that will update collected `samples`s' `labels` with `self._id` or something alike.
The ID will be different if is YARN or not obviously, as now.

roi-granulate (approved):
LGTM

granulate_utils/metrics/yarn.py
```diff
@@ -16,10 +17,14 @@
 
 class ResourceManagerAPI:
     def __init__(self, rm_address: str):
+        self._info_url = f"{rm_address}/ws/v1/cluster/info"
```
Jongy, line 20: Why is it added here and also in YARN_CLUSTER_INFO_PATH? (same applies to other paths here I see..)

granulate_utils/metrics/sampler.py
```diff
@@ -263,12 +266,18 @@ def _init_collectors(self):
...
+            self._collectors.append(yarn_collector)
+            self._cluster_id = self._cluster_id or yarn_collector.cluster_id
 
-        # In Standalone and Mesos we'd use applications metrics
-        if self._cluster_mode in (SPARK_STANDALONE_MODE, SPARK_MESOS_MODE):
+        elif self._cluster_mode in (SPARK_STANDALONE_MODE, SPARK_MESOS_MODE):
+            # In Standalone and Mesos we'd use applications metrics
             self._applications_metrics = True
 
+            # Didn't research how to bring the actual cluster id, so use the master address
```
Jongy, line 277: Since we add cluster_id only for YARN, and for SA we remain with the master_address which is converted to "hostname" and is available in our queries - there's no purpose to add it as a label for SA and instead of having the rather patchy `_add_cluster_id_label`, you can add the YARN cluster_id inside `YarnCollector` to all metrics added by `YarnCollector`.

Jongy (changes requested):

granulate_utils/metrics/sampler.py
roi-granulate: note your suggestion does not support YARN with `applications_metrics` enabled

roi-granulate: I still think that having `Sampler` id is a good practice
If you find `_add_cluster_id_label` problematic or risking being high resource consumption, we might want to somehow transfer to the collectors `cluster_labels` which will be added by the collectors themselves while metrics are collected

roi-granulate: I still think that having `Sampler` id is a good practice
If you find `_add_cluster_id_label` problematic, we might want to somehow transfer to the collectors `cluster_labels` which will be added by the collectors themselves while metrics are collected

Jongy:
Closing this PR as the metrics package was removed in #220 to a separate repository. Reopen there if relevant.

PR #167, opened 2023-07-09
Fetch AWS ECS container metadata

d3dave:
Tested on Fargate.

```yaml
container_arn: arn:aws:ecs:us-east-1:XXXXXX:container/...
execution_env: AWS_ECS_FARGATE
provider: aws
region: us-east-1
```

### DIFF
--- a/granulate_utils/metadata/cloud.py
+++ b/granulate_utils/metadata/cloud.py
@@ -103,15 +103,15 @@ def get_aws_instance_metadata() -> Optional[AwsInstanceMetadata]:
 
 
 def get_aws_container_metadata() -> Optional[AwsContainerMetadata]:
-    ecs_container_metadata_uri_v4 = os.environ['ECS_CONTAINER_METADATA_URI_V4']
+    ecs_container_metadata_uri_v4 = os.environ["ECS_CONTAINER_METADATA_URI_V4"]
     response = send_request(ecs_container_metadata_uri_v4)
     if response is None:
         return None
     metadata = response.json()
     return AwsContainerMetadata(
         provider="aws",
-        execution_env=os.environ['AWS_EXECUTION_ENV'],
-        region=os.environ['AWS_REGION'],
+        execution_env=os.environ["AWS_EXECUTION_ENV"],
+        region=os.environ["AWS_REGION"],
         container_arn=metadata["ContainerARN"],
     )
 
--- a/granulate_utils/metadata/cloud.py
+++ b/granulate_utils/metadata/cloud.py
@@ -22,12 +22,12 @@ METADATA_REQUEST_TIMEOUT = 5
 
 
 @dataclass
-class InstanceMetadataBase:
+class CloudMetadataBase:
     provider: str
 
 
 @dataclass
-class AwsInstanceMetadata(InstanceMetadataBase):
+class AwsInstanceMetadata(CloudMetadataBase):
     region: str
     zone: str
     instance_type: str
@@ -38,7 +38,14 @@ class AwsInstanceMetadata(InstanceMetadataBase):
 
 
 @dataclass
-class GcpInstanceMetadata(InstanceMetadataBase):
+class AwsContainerMetadata(CloudMetadataBase):
+    execution_env: str
+    region: str
+    container_arn: str
+
+
+@dataclass
+class GcpInstanceMetadata(CloudMetadataBase):
     provider: str
     zone: str
     instance_type: str
@@ -50,7 +57,7 @@ class GcpInstanceMetadata(InstanceMetadataBase):
 
 
 @dataclass
-class AzureInstanceMetadata(InstanceMetadataBase):
+class AzureInstanceMetadata(CloudMetadataBase):
     provider: str
     instance_type: str
     zone: str
@@ -63,7 +70,7 @@ class AzureInstanceMetadata(InstanceMetadataBase):
     image_info: Optional[Dict[str, str]]
 
 
-def get_aws_metadata() -> Optional[AwsInstanceMetadata]:
+def get_aws_instance_metadata() -> Optional[AwsInstanceMetadata]:
     # Documentation:
     # on the format: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
     # on the protocol: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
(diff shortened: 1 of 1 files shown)


d3dave:
Ok, this isn't good enough. I just tested on plain ECS (`AWS_EXECUTION_ENV=AWS_ECS_EC2`) and we get `KeyError: 'AWS_REGION'` :(

There we do have a container metadata endpoint as well. There is no dedicated region field. The region is present only within the Container/Task ARN. Considering ARN format is very well-defined, I suppose we should just pare the region from it: `arn:partition:service:region:account-id:resource`

granulate_utils/metadata/cloud.py
```diff
@@ -95,6 +102,20 @@ def get_aws_metadata() -> Optional[AwsInstanceMetadata]:
     )
 
 
+def get_aws_container_metadata() -> Optional[AwsContainerMetadata]:
+    ecs_container_metadata_uri_v4 = os.environ["ECS_CONTAINER_METADATA_URI_V4"]
```
Jongy, line 106:
```suggestion
    ecs_container_metadata_uri_v4 = os.environ.get("ECS_CONTAINER_METADATA_URI_V4")
    if ecs_container_metadata_uri_v4 is None:
        return None
```

Jongy (changes requested):
One small comment.

> Ok, this isn't good enough. I just tested on plain ECS (AWS_EXECUTION_ENV=AWS_ECS_EC2) and we get KeyError: 'AWS_REGION' :(
> There we do have a container metadata endpoint as well. There is no dedicated region field. The region is present only within the Container/Task ARN. Considering ARN format is very well-defined, I suppose we should just parse the region from it: arn:partition:service:region:account-id:resource

In either case, we get the region from the instance metadata, don't we?

granulate_utils/metadata/cloud.py
```diff
@@ -165,9 +186,14 @@ def send_request(url: str, headers: Dict[str, str] = None, method: str = "get")
     return response
 
 
-def get_static_cloud_instance_metadata(logger: Union[logging.LoggerAdapter, logging.Logger]) -> Optional[Metadata]:
+def get_static_cloud_metadata(logger: Union[logging.LoggerAdapter, logging.Logger]) -> Optional[Metadata]:
```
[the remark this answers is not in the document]
Jongy: This break is a simple fix & cannot go over unnoticed, so I see no reason to spend time on deprecation etc.

d3dave:
> One small comment.
> 
> > Ok, this isn't good enough. I just tested on plain ECS (AWS_EXECUTION_ENV=AWS_ECS_EC2) and we get KeyError: 'AWS_REGION' :(
> > There we do have a container metadata endpoint as well. There is no dedicated region field. The region is present only within the Container/Task ARN. Considering ARN format is very well-defined, I suppose we should just parse the region from it: arn:partition:service:region:account-id:resource
> 
> In either case, we get the region from the instance metadata, don't we?

Yes. Just for the sake of completeness if one wishes to get both. But I suppose it's not important.

granulate_utils/metadata/cloud.py
```diff
@@ -95,6 +102,22 @@ def get_aws_metadata() -> Optional[AwsInstanceMetadata]:
...
+    if ecs_container_metadata_uri_v4 is None:
+        return None
+    response = send_request(ecs_container_metadata_uri_v4)
+    if response is None:
+        return None
+    metadata = response.json()
+    return AwsContainerMetadata(
+        provider="aws",
+        execution_env=os.environ["AWS_EXECUTION_ENV"],
+        region=os.environ["AWS_REGION"],
```
Jongy, line 116: Make it optional then?

Jongy (changes requested):
Let's make it optional then

granulate_utils/metadata/cloud.py
d3dave: the existence of this envar is optional, but we always need to know the region

d3dave: empirically it exists on Fargate, which is what we are currently concerned with. are we concerned with non-Fargate environments?

Jongy: We are concerned with region, but we get it from the instance metadata, i.e `get_aws_instance_metadata`.

This raises a problem with your design - `get_static_cloud_metadata` returns the first match, and only it, because it segregates based on providers. What you're adding here is additional types of metadata that we want to be collected together & merged & presented as a single result to the backend.

Practically speaking, I suggest we currently add the `execution_env` in ALL cases to the AWS metadata we return. And specifically for fargate, we will send the request to ECS_CONTAINER_METADATA_URI_V4 to get the container ARN, and add it to the metadata as well. In other cases (all non-fargate) it can be none.
Please check if instance metadata is available in fargate - if it does then we do not need to concern ourselves with getting the region from additional sources.

Jongy (changes requested):
.

granulate_utils/metadata/cloud.py
d3dave: instance metadata is not available in fargate, because tasks are not running on an instance. they run in a dedicated vm

Jongy: So in fargate - the instance metadata fetcher will fail, but in ECS, but fetchers will succeed and arbitrarily one of them will be returned. Therefore please merge them to a single one as I suggested.

d3dave: what do you mean arbitrarily?

Jongy: > get_static_cloud_metadata returns the first match, and only it, because it segregates based on providers

It calls all fetchers in parallel and returns the first one that has a result.

d3dave: yes. does gcp or azure metadata work on ecs?

Jongy: Discussed f2f - `get_fargate_metadata`

d3dave: @Jongy done

granulate_utils/metadata/cloud.py
```diff
@@ -38,7 +38,14 @@ class AwsInstanceMetadata(InstanceMetadataBase):
 
 
 @dataclass
-class GcpInstanceMetadata(InstanceMetadataBase):
+class AwsContainerMetadata(CloudMetadataBase):
```
Jongy, line 41: This should have `provider: aws` as well.

d3dave: it is initialized with provider="aws". do you mean set as default `provider: str = 'aws'`?

Jongy: We have the provider field in `AwsInstanceMetadata` but not in `AwsContainerMetadata`

d3dave: image

we have the provider field everywhere. its not explicitly mentioned in AwsInstanceMetadata either, and it doesn't need to be. it is unfortunate that we have the inconsistency of it being mentioned in Gcp... and Azure... even when they inherit it from the base class as well.

Jongy: Oof, sorry, I looked at GCP and Azure and not AWS so it confused me. You can fix GCP and Azure if you'd like on this occasion.

d3dave: I haven't decided myself whether it's better for it to be mentioned and set to a default, or just inherited and not mentioned 🤔 
Perhaps we can revisit the question in the future

Jongy (approved):
Approved, one fine note though - maybe add execution_env to `AwsInstanceMetadata` as well? To simplify the is_fargate check later. Or you just plan to check if the field execution_env exists?

PR #168, opened 2023-07-10
[config-feeder] add autoscaling config collector

slicklash:
add autoscaling config collector to config feeder


Jongy:
Moved out.

PR #170, opened 2023-07-12
Databricks: match better on the regex that trims run-id

DanielShaulov:
This changes the "-" that is removed to be the one before the "run-%d" instead of the one afterwards to catch cases where the cluster name is just "job-%d-run-%d"


roi-granulate (approved):
LGTM

PR #173, opened 2023-07-18
introduce external collectors

roymiloh:
The ability to pass collectors from outside, with a preconfigured logger.

I chose to pass a collection of factories, so the caller will be able to pass parameters upon collector creation, e.g.

```python
class MyCollector(ConfigFeederCollector)
  name = "my_collector"

  def collect(self, node_info):
    return CollectionResult(config=...)

def my_collector(params):
  # Here the caller may pass some custom params
  return MyCollector(params, ...)

ConfigFeederClient(..., collector_factories=[my_collector])
```


granulate_utils/config_feeder/client/__init__.py
```diff
@@ -1,5 +1,13 @@
...
+from granulate_utils.config_feeder.client.collector import ConfigFeederCollector
 
-__all__ = ["ConfigFeederClient", "ClientError", "APIError", "MaximumRetriesExceeded", "get_node_info"]
+__all__ = [
+    "ConfigFeederClient",
+    "ClientError",
+    "APIError",
+    "MaximumRetriesExceeded",
+    "ConfigFeederCollector",
+    "get_node_info",
```
roymiloh, line 12: Should I expose `HttpClient` as well? for typing purposes (to be used by external collectors).

slicklash: maybe remove ConfigFeederCollector from here? these exports are for bare minimal usage. advanced users can import  directly from `granulate_utils.config_feeder.client.collector`

granulate_utils/config_feeder/client/http_client.py
```diff
@@ -0,0 +1,47 @@
...
+
+from granulate_utils.config_feeder.client.exceptions import APIError, ClientError
+from granulate_utils.config_feeder.core.errors import raise_for_code
+
+DEFAULT_API_SERVER_ADDRESS = "https://api.granulate.io/config-feeder/api/v1"
+DEFAULT_REQUEST_TIMEOUT = 3
+
+
+class HttpClient:
+    def __init__(self, token: str, server_address: Optional[str]) -> None:
```
slicklash, line 15: replace `token` with `headers: Dict[str, str]` here, this will allow collectors to pass custom headers when needed.

roymiloh: I actually like it when it's explicit, as it (token) is a mandatory param (better business-wise reflection). We can add additional custom params if needed later.

slicklash:
Since HttpClient is same for all collectors ConfigFeederClient should handle cluster / node registration and submission.
Collectors should do only collections and config feeder should handle talking with API

payload structure could be

```
{
  "collector_name": "<collection_result>",
}

e.g.

{
    "yarn_config": ...,
    "autoscaling_config": ...,
    "custom1": ...,
}
```
one request to submit everything to server. If there is a need to talk with server during collection http client still can be provided to collectors

granulate_utils/config_feeder/client/collector.py
```diff
@@ -0,0 +1,27 @@
...
+
+class ConfigFeederCollector(ABC):
+    def __init__(self, params: ConfigFeederCollectorParams) -> None:
+        self.logger = params["logger"]
+        self._http_client = params["http_client"]
+        self._service = params["service"]
+        self._collector_type = params["collector_type"]
+
+    @abstractmethod
+    async def collect(self) -> None:
```
slicklash, line 26: most of collectors will need node_info
```python
async def collect(self, node_info: NodeInfo) -> Any:
```

collerek: Should we return `Dicts` or `BaseModels` from `collect`?

roymiloh: I think a generic `dict` is fine.

granulate_utils/config_feeder/client/cluster_client.py
```diff
@@ -0,0 +1,71 @@
...
+            self._register_cluster(node_info)
+
+    def submit_node_configs(
+        self,
+        node_info: NodeInfo,
+        config_requests: Dict[str, NodeResourceConfigCreate],
+    ) -> Dict[str, Any]:
+        node_id = self._register_node(node_info)
+        self._logger.info(f"sending configs for node {node_info.external_id}")
+        request = {}
```
roymiloh, line 33: Not great, but not terrible - the alternative is to somehow use Pydantic's `create_model` and create this model in runtime, not sure it's worth it.

granulate_utils/config_feeder/client/http_client.py
roymiloh: `HttpClient` is not longer passed as a param to the collectors. If it's needed, I'll consider this.

granulate_utils/config_feeder/client/cluster_client.py
```diff
@@ -0,0 +1,71 @@
...
+
+from granulate_utils.config_feeder.client.http_client import HttpClient
+from granulate_utils.config_feeder.client.models import CollectionResult
+from granulate_utils.config_feeder.core.models.aggregation import NodeResourceConfigCreate
+from granulate_utils.config_feeder.core.models.cluster import ClusterCreate, CreateClusterRequest, CreateClusterResponse
+from granulate_utils.config_feeder.core.models.collection import CollectorType
+from granulate_utils.config_feeder.core.models.node import CreateNodeRequest, CreateNodeResponse, NodeCreate, NodeInfo
+
+
+@dataclass
```
slicklash, line 14: nit pick: not fan of data classes doing I/O operations

slicklash (approved):

PR #179, opened 2023-08-04
Add identification of golang binaries

marcin-ol:
Support identification of Golang binaries by looking for ELF section `.note.go.buildid`, supported since Go 1.5.
Reference: https://github.com/golang/go/issues/11048.

### DIFF
--- a/granulate_utils/golang.py
+++ b/granulate_utils/golang.py
@@ -25,8 +25,7 @@ def get_golang_buildid(process: Process) -> Optional[str]:
     elf_path = f"/proc/{process.pid}/exe"
     try:
         # section .note.go.buildid has been added since version 1.5: https://github.com/golang/go/issues/11048
-        golang_buildid = get_elf_buildid(elf_path, ".note.go.buildid", lambda note: note.n_name == "Go")
-        return golang_buildid
+        return get_elf_buildid(elf_path, ".note.go.buildid", lambda note: note.n_name == "Go")
     except FileNotFoundError:
         raise NoSuchProcess(process.pid)
 
--- a/granulate_utils/golang.py
+++ b/granulate_utils/golang.py
@@ -9,12 +9,26 @@ from typing import Optional
 
 from psutil import NoSuchProcess, Process
 
-from granulate_utils.linux.elf import read_elf_symbol, read_elf_va
+from granulate_utils.linux.elf import get_elf_buildid, read_elf_symbol, read_elf_va
 from granulate_utils.linux.process import is_kernel_thread
 
 
 def is_golang_process(process: Process) -> bool:
-    return not is_kernel_thread(process) and get_process_golang_version(process) is not None
+    return not is_kernel_thread(process) and get_golang_buildid(process) is not None
+
+
+@functools.lru_cache(maxsize=4096)
+def get_golang_buildid(process: Process) -> Optional[str]:
+    """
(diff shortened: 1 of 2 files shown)


granulate_utils/golang.py
```diff
@@ -9,12 +9,26 @@
...
+@functools.lru_cache(maxsize=4096)
+def get_golang_buildid(process: Process) -> Optional[str]:
+    """
+    Gets the golang build ID embedded in an ELF file section as a string, or None if not present.
+    """
+    elf_path = f"/proc/{process.pid}/exe"
+    try:
+        # section .note.go.buildid has been added since version 1.5: https://github.com/golang/go/issues/11048
+        golang_buildid = get_elf_buildid(elf_path, ".note.go.buildid", lambda note: note.n_name == "Go")
+        return golang_buildid
```
Jongy, line 29:
```suggestion
        return get_elf_buildid(elf_path, ".note.go.buildid", lambda note: note.n_name == "Go")
```

granulate_utils/linux/elf.py
```diff
@@ -71,18 +71,19 @@ def get_elf_arch(elf: ELFType) -> str:
         return elf.get_machine_arch()
 
 
-def get_elf_buildid(elf: ELFType) -> Optional[str]:
+def get_elf_buildid(elf: ELFType, section: str, note_check: Callable[[Any], Any] = lambda n: True) -> Optional[str]:
```
Jongy, line 74: Does it make sense to have a default value for the note? And furthermore, a default value that behaves differently that the past behavior?

I'd just remove the default and require the caller to pass it.

If you can add types that's also good :)

Jongy:
Related: https://github.com/Granulate/gprofiler/issues/817

granulate_utils/golang.py
marcin-ol: 👍

granulate_utils/linux/elf.py
marcin-ol: 👍

Jongy:
Btw this change will have a side effect of greatly improving the performance (iterating over sections is much less work than iterating over symbols), helping for https://github.com/Granulate/gprofiler/issues/696 cause.

granulate_utils/linux/elf.py
```diff
@@ -71,18 +71,19 @@ def get_elf_arch(elf: ELFType) -> str:
...
+def get_elf_buildid(elf: ELFType, section: str, note_check: Callable[[NoteSection], bool]) -> Optional[str]:
     """
-    Gets the build ID embedded in an ELF file section as a hex string,
+    Gets the build ID embedded in an ELF file note section as a hex string,
     or None if not present.
+    Lambda argument is used to verify that note meets caller's requirements.
     """
     with open_elf(elf) as elf:
-        build_id_section = elf.get_section_by_name(".note.gnu.build-id")
+        build_id_section = elf.get_section_by_name(section)
```
Jongy, line 81: Can name it note_section as it's not necessarily a build-id one anymore

Jongy (changes requested):
One final small comment.

Other than that I'll have to review the usage of changed functions (`is_golang_process` and `get_elf_buildid`) in other project which use granulate-utils to ensure compatibility is kept.

Jongy:
No problematic usages. I'm ready to merge this - just waiting on your reply on the last comment.

Jongy (approved):

PR #180, opened 2023-08-06
linux: ekf: Use raise_nosuchprocess on open_elf() so it catches all

Jongy:
@d3dave is this the right way or should I swap contextmanager and raise_nosuchprocess?


d3dave:
> @d3dave is this the right way or should I swap contextmanager and raise_nosuchprocess?

Neither. The `yield ...` or `with ...` must be entirely within `try-except` which you can't do like this. Simplest approach is probably to swap the decorators (so you don't have to define the `raise` func as a `contextmanager`) and then to do a `yield from` if the wrapped function is a generator function or if the value is a generator.

d3dave (approved):

PR #186, opened 2023-08-16
Java: fix `parse_jvm_version` `vm_name` regex pattern

roi-granulate:
1. better regex:
e.g. this regex was failed to match the [functions' own `java -version` output example](https://github.com/Granulate/granulate-utils/blob/dadafab/granulate_utils/java.py#L136)
2. on my PC it's `Java HotSpot(TM) 64-Bit Server VM`... so I added support
3. replaced `TODO` with an informative comment
4. tests, including the documented `java -version` example from `parse_jvm_version()`

one question I have is why do we need the regex? why can't we just `lines[2].startswith()` or even `if 'HotSpot' in lines[2]`?

### DIFF
--- a/granulate_utils/java.py
+++ b/granulate_utils/java.py
@@ -191,18 +191,18 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
         build = int(matched[0])
 
     # There is no real format here, just use the entire description string until the part which describes the build.
-    m = re.match(r"(.*?) \((?:product )?build", lines[2])
+    m = re.match(r"(.*?) (?:\(.*\))?\((?:product )?build", lines[2])
     assert m is not None, f"Missing build description? {lines[2]}"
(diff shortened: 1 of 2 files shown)


Jongy (approved):
You can merge, and create a PR in gProfiler that updates the granulate-utils reference.

Jongy:
> better regex:
> e.g. this regex was failed to match the [functions' own java -version output example](https://github.com/Granulate/granulate-utils/blob/dadafab/granulate_utils/java.py#L136)

You can add the example as a testcase as well :)

roi-granulate:
@Jongy 

> You can add the example as a testcase as well :)


> 4. tests, including the documented java -version example from parse_jvm_version()

Jongy:
Good @roi-granulate I missed it. Thanks

PR #187, opened 2023-08-16
Support Mappings in the Metadata type

netaneld122:
Support Mappings in Metadata so we could pass other dict-like objects.


netaneld122:
Breaks too much stuff, closing

PR #190, opened 2023-08-24
fix: hadoop v2.6.0 compatibility

slicklash:
HADOOP v2 and v3 behaves differently. In hadoop v2 nested JSON objects can contain nulls.

e.g. `curl http://localhost:8088/ws/v1/cluster/apps?finalStatus=KILLED`
v3 response: `{"apps":{}}`
v2 response: `{"apps":null}`

### DIFF
--- a/granulate_utils/metrics/yarn.py
+++ b/granulate_utils/metrics/yarn.py
@@ -6,7 +6,10 @@
 # Licensed under a 3-clause BSD style license (see LICENSE.bsd3).
 #
 import logging
+from functools import cached_property
 from typing import Dict, Iterable, List, Optional
+
+from packaging.version import Version
 
 from granulate_utils.metrics import Collector, Sample, json_request, samples_from_json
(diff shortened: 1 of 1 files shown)


granulate_utils/metrics/yarn.py
```diff
@@ -6,8 +6,11 @@
 # Licensed under a 3-clause BSD style license (see LICENSE.bsd3).
 #
 import logging
+from functools import cached_property
 from typing import Dict, Iterable, List, Optional
 
+from packaging.version import Version
```
d3dave, line 12: `packaging` is not a builtin, it must be installed separately. Either add to requirements or do smth else.

d3dave (changes requested):

granulate_utils/metrics/yarn.py
slicklash: It's already there: https://github.com/Granulate/granulate-utils/blob/823c27ef6d87f9358e6b321da1919bb66daa5d4f/requirements.txt#L10

d3dave (approved):

PR #192, opened 2023-08-27
Make python pre-commit hook fail if it would reformat

michelhe:
Currently, when I run `git commit` with a staged file that would be reformatted by the pre-commit hook, the hook will reformat the file but the the commit succeed with the unforamtted staged files.

I'd like to make the pre-commit hook fail if it would reformat the file, so that I can fix the formatting and then commit.

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -16,6 +16,6 @@ fi
 EXCLUDE_RE='.venv|venv|build|granulate_utils/generated'
 
 python3 -m isort --settings-path .isort.cfg $check_arg --skip granulate_utils/generated --skip venv --skip .venv --skip build .
-python3 -m black --line-length 120 $check_arg --exclude $EXCLUDE_RE .
+python3 -m black --check --diff --color --line-length 120 $check_arg --exclude $EXCLUDE_RE .
 python3 -m flake8 --config .flake8  --exclude $(echo $EXCLUDE_RE | tr '|' ',') .
 python3 -m mypy --exclude $EXCLUDE_RE .


lint.sh
```diff
@@ -16,6 +16,6 @@ fi
 EXCLUDE_RE='.venv|venv|build|granulate_utils/generated'
 
 python3 -m isort --settings-path .isort.cfg $check_arg --skip granulate_utils/generated --skip venv --skip .venv --skip build .
-python3 -m black --line-length 120 $check_arg --exclude $EXCLUDE_RE .
+python3 -m black --check --diff --color --line-length 120 $check_arg --exclude $EXCLUDE_RE .
```
Jongy, line 19: You need to do some `if ci` here so when running locally it does fix things.
In gProfiler it's done this way:
```
black_extra_args=""
isort_extra_args=""
if [[ "$1" = "--ci" ]]; then
    black_extra_args="--check"
    isort_extra_args="--check-only"
fi
```

You can fix it  in gProfiler as well

michelhe: I also want it to `--check` in the precommit hook.

You imagine some argument to `lint.sh` that does either `--check` or `--fix` ?

Jongy: Currently it edits in place or only gives diff?

michelhe: It edits in-place, but returns 0 so you actually succeed in the hook and get the commit editmsg before the reformatted stuff is staged

Jongy: Okie

Jongy (approved):

lint.sh
michelhe: My above statement refers to the "current" state in master.

My change causes it to no longer reformat, just display diff and exit 1, yeah?

Jongy: No. Default behavior is to format. In ci pass a flag and only show diff and exit

Jongy (changes requested):
.

lint.sh
michelhe: But the point I'm arguing at is that also in development I don't want the default behavior.
What good is a pre-commit hook if it reformats but then let's you commit the non-formatted code?

Jongy: Dude what do you suggest we're nitpicking about this small thing for the entire day lol. If black doesn't fix for u in the script, then u need to manually run black with the same flags which is not convenient. The script needs to fix it for you. I understand a problem exists now but your suggestion worsens the UX

PR #195, opened 2023-09-06
databricks client: use exceptions and add logging, SparkUI requests control

guybortnikov:
This pull request improves databricks metadata client, adds more logging, and adds the ability to control its defaults.
Also, it adds support to control parameters when accessing SparkUI for metrics.

### DIFF
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -127,7 +127,7 @@ class DBXWebUIEnvWrapper:
 
         if self.raise_on_failure:
             raise DatabricksMetadataFetchException(
-                "Failed to get DBX environment metadata in timeout " f"of {DATABRICKS_JOBNAME_TIMEOUT_S} seconds"
+                f"Timed out getting DBX environment metadata after {DATABRICKS_JOBNAME_TIMEOUT_S}s"
             )
         self.logger.info("Databricks get DBX environment metadata timeout reached")
         return None
--- a/granulate_utils/metadata/databricks_client.py
+++ b/granulate_utils/metadata/databricks_client.py
@@ -12,7 +12,14 @@ from typing import Any, Dict, Optional
 
 import requests
 
-from granulate_utils.exceptions import DatabricksJobNameDiscoverException
+from granulate_utils.exceptions import (
+    DatabricksJobNameDiscoverException,
+    DatabricksMetadataFetchException,
+    DatabricksTagsExtractionException,
+    SparkAPIException,
+    SparkAppsURLDiscoveryException,
+    SparkNotReadyException,
+)
 
 HOST_KEY_NAME = "*.sink.ganglia.host"
 DATABRICKS_METRICS_PROP_PATH = "/databricks/spark/conf/metrics.properties"
@@ -35,8 +42,10 @@ CLUSTER_USAGE_RELEVANT_TAGS_PROPS = [
     "spark.databricks.clusterUsageTags.driverNodeType",
 ]
 DATABRICKS_REDACTED_STR = "redacted"
-SPARKUI_APPS_URL = "http://{}/api/v1/applications"
-REQUEST_TIMEOUT = 5
+SPARKUI_APPS_URL_FORMAT = "http://{}/api/v1/applications"
+DEFAULT_REQUEST_TIMEOUT = 15
+DEFAULT_REQUEST_VERIFY_SSL = False
+
 JOB_NAME_KEY = "RunName"
 CLUSTER_NAME_KEY = "ClusterName"
 DEFAULT_WEBUI_PORT = 40001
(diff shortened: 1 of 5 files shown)


granulate_utils/metrics/spark.py
```diff
@@ -268,7 +283,12 @@ def _spark_executor_metrics(self, running_apps: Dict[str, Tuple[str, str]]) -> I
         for app_id, (app_name, tracking_url) in running_apps.items():
             try:
                 base_url = get_request_url(self.master_address, tracking_url)
-                executors = rest_request_to_json(base_url, SPARK_APPS_PATH, app_id, "executors")
+                executors = rest_request_to_json(
+                    base_url,
+                    SPARK_APPS_PATH,
+                    app_id,
+                    "executors",
+                )
```
d3dave, line 291: revert this

granulate_utils/metadata/databricks_client.py
```diff
@@ -46,12 +55,25 @@
...
+        self,
+        logger: logging.LoggerAdapter,
+        enable_retries: bool = True,
+        http_request_timeout: int = DEFAULT_REQUEST_TIMEOUT,
+        raise_on_failure: bool = False,
+        verify_request_ssl: bool = DEFAULT_REQUEST_VERIFY_SSL,
+    ) -> None:
         """
         When `enable_retries` is True, the wrapper will retry the request to the webui until it succeeds or until
         """
```
d3dave, line 68: The sentence here is cut off 😕

granulate_utils/metadata/databricks_client.py
```diff
@@ -85,24 +115,25 @@ def extract_relevant_metadata(self) -> Optional[Dict[str, str]]:
...
-        self.logger.info("Databricks get DBX environment metadata timeout, continuing...")
+
+            # No environment metadata yet, retry.
+            time.sleep(RETRY_INTERVAL_S)
+
+        self.logger.info("Databricks get DBX environment metadata timeout reached")
+
+        if self.raise_on_failure:
+            raise DatabricksMetadataFetchException(
+                "Failed to get DBX environment metadata in timeout " f"of {DATABRICKS_JOBNAME_TIMEOUT_S} seconds"
```
d3dave, line 135:
```suggestion
                f"Timed out getting DBX environment metadata after {DATABRICKS_JOBNAME_TIMEOUT_S}s"
```

granulate_utils/metadata/databricks_client.py
```diff
@@ -85,24 +115,25 @@ def extract_relevant_metadata(self) -> Optional[Dict[str, str]]:
...
+                    raise e
+
             if not self.enable_retries:
                 break
-        self.logger.info("Databricks get DBX environment metadata timeout, continuing...")
+
+            # No environment metadata yet, retry.
+            time.sleep(RETRY_INTERVAL_S)
+
+        self.logger.info("Databricks get DBX environment metadata timeout reached")
```
d3dave, line 131: This message looks redundant if we are raising the exception below. Maybe we should only log this if we do not raise?

granulate_utils/metadata/databricks_client.py
```diff
@@ -115,109 +146,107 @@ def _discover_apps_url(self) -> bool:
...
         """
-        Returns `includes spark.databricks.clusterUsageTags.clusterAllTags` tags as `Dict`.
+        Returns `includes spark.databricks.clusterUsageTags.clusterAllTags` tags as `dict`.
         In any case this function returns `None`, a retry is required.
         """
         if not os.path.isfile(DATABRICKS_METRICS_PROP_PATH):
+            self.logger.warning(f"{DATABRICKS_METRICS_PROP_PATH} was not found.")
             # We want to retry in case the cluster is still initializing, and the file is not yet deployed.
-            return None
+            raise SparkAppsURLDiscoveryException(f"{DATABRICKS_METRICS_PROP_PATH} was not found.")
```
d3dave, line 181: Logging messages before exceptions looks redundant. All the exceptions are always logged by the caller anyway afaict. If the log level is important, then we can add a value to the exception classes that says which level they should be logged at.

d3dave (changes requested):

granulate_utils/metadata/databricks_client.py
d3dave: Also needs doc-ing for the new values.

granulate_utils/metrics/spark.py
guybortnikov: This is the git pre-commit hook, why revert? It's just alignment

granulate_utils/metadata/databricks_client.py
guybortnikov: Old @roi-granulate's code, removing the comment

granulate_utils/metadata/databricks_client.py
guybortnikov: I agree but I'm afraid of a situation where I'm unaware of all the callers, this is why I also added an argument indicating whether to raise or not.  
It's either:
a. We drop all the log messages
b. We'll log only if not raise arg is opted in

granulate_utils/metrics/spark.py
d3dave: because it looks worse than the alternative

d3dave (approved):

granulate_utils/metrics/spark.py
Jongy: @d3dave agree but it's the autoformatter, I'm not blocking @guybortnikov on this, we can fix the format later.

PR #199, opened 2023-09-10
Glogger extra exception

michelhe:
This PR adds the ability for `ExtraAdapter` to collect `extra` attrs from Exception objects, for use when doing:
`adapater.exception("An exception has occured")`

For convenience, also added an `ExtraException` base-class

### DIFF
--- a/glogger/extra_adapter.py
+++ b/glogger/extra_adapter.py
@@ -1,3 +1,7 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
 import logging
 from typing import Any, Dict, Mapping, MutableMapping, Tuple
 
--- a/glogger/extra_adapter.py
+++ b/glogger/extra_adapter.py
@@ -6,8 +6,6 @@ import logging
 import sys
 from typing import Any, Dict, Mapping, MutableMapping, Tuple
 
-from .extra_exception import ExtraException
-
 
 class ExtraAdapter(logging.LoggerAdapter):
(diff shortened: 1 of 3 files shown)


glogger/extra_adapter.py
```diff
@@ -39,6 +46,17 @@ def process(self, msg: Any, kwargs: MutableMapping[str, Any]) -> Tuple[Any, Muta
...
 
         extra = self.get_extra(**logging_kwargs)
+
+        if logging_kwargs.get("exc_info") is True:
+            # If exc_info is True, and the exception is subclassing ExtraException, then add the extra attributes
+            # from the exception to the extra attributes of the record:
+            exc_info = sys.exc_info()
+            # If exc_info is True, then there must be an exception:
+            assert exc_info is not None
+            if isinstance(exc_info[1], ExtraException):
```
d3dave, line 56: Don't actually have to require it. I don't think logging does. Also, I don't want to derive from `ExtraException` everywhere now. It's easier to just support it transparently by checking for `.extra`, or even better can just `extra = e.__dict__`
```suggestion
            if exc_info and hasattr(exc_info[1], 'extra'):
```

Jongy: I don't think we want it implicit (i.e ALWAYS take the `__dict__` of an exception). I agree it's nicer without requiring to subclass `ExtraException` tho.

michelhe: So I'm keeping the base class for syntactic sugar and having the adapter doing getattr("extra"), aye?

Jongy (approved):

PR #212, opened 2023-11-30
Added new method to get yarn home dir

LiorMosk:


granulate_utils/metrics/yarn/utils.py
```diff
@@ -165,10 +167,26 @@ def find_yarn_home_dir(*, logger: Union[logging.Logger, logging.LoggerAdapter])
     Find YARN home directory from command line arguments
     """
     logger.debug("looking for running YARN processes")
-    lines = subprocess.run(["ps", "-ax"], capture_output=True, text=True).stdout.split(" -D")
+
+    lines = subprocess.run(["ps", "-ax"], capture_output=True, text=True).stdout
```
LiorMosk, line 171: We can maybe filter only 'yarn' user processes but I'm not sure that this user is exists in every yarn environment

Jongy: Not sure it exists + not sure that in the context this code executes, we'll be able to resolve the numeric UID to the string UID `yarn`. So no need to do that.

granulate_utils/metrics/yarn/utils.py
```diff
@@ -165,10 +167,26 @@ def find_yarn_home_dir(*, logger: Union[logging.Logger, logging.LoggerAdapter])
...
+    processes_records = lines.splitlines()
+    for process_rec in processes_records:
+        if "yarn" in process_rec:  # Assuming if 'yarn' in cmd, it is a yarn process.
+            try:
+                process_pid = int(process_rec.split(" ")[0])
+                process = Process(process_pid)
+                for env_key, env_val in process.environ().items():
+                    if "HADOOP_YARN_HOME" == env_key:
```
Jongy, line 179: Constify this HADOOP_YARN_HOME

Jongy (changes requested):
Done reviewing

Jongy (approved):

tests/granulate_utils/metrics/fixtures.py
```diff
@@ -85,6 +86,9 @@ def _mock_subprocess_run_stdout(self: T, *args: Any, **kwargs: Any) -> Mock:
         self._mock.stdout = self._stdout[cmd]
         return self._mock
 
+    def mock_path_is_dir(self) -> None:
+        self._contexts.add(patch("pathlib.Path.is_dir", lambda path: True))
```
Jongy, line 90: Mocking all dirs doesn't make sense to me. This can create problems or hide problems. Mock only the particular things you need

PR #213, opened 2023-12-03
fix detect_yarn_config finding yarn-site method

LiorMosk:


granulate_utils/metrics/yarn/utils.py
```diff
@@ -158,8 +159,14 @@ def detect_yarn_config(*, logger: Union[logging.Logger, logging.LoggerAdapter])
     """
     if yarn_home_dir := find_yarn_home_dir(logger=logger):
         logger.debug(f"found YARN home dir: {yarn_home_dir}")
-        yarn_site_xml_file = Path(yarn_home_dir).joinpath(RELATIVE_YARN_SITE_XML_PATH)
-        return read_config_file(yarn_site_xml_file, logger=logger)
+        try:
+            yarn_site_xml_file = glob.glob(f"{yarn_home_dir}/**/yarn-site.xml", recursive=True)[0]
```
Jongy, line 163: Log a warning if more than one is found, this is not expected and may cause errors.

granulate_utils/metrics/yarn/utils.py
```diff
@@ -160,6 +162,14 @@ def detect_yarn_config(*, logger: Union[logging.Logger, logging.LoggerAdapter])
         logger.debug(f"found YARN home dir: {yarn_home_dir}")
         yarn_site_xml_file = Path(yarn_home_dir).joinpath(RELATIVE_YARN_SITE_XML_PATH)
         return read_config_file(yarn_site_xml_file, logger=logger)
+
+    # try to find YARN config in environment variable
+    else:
```
Jongy, line 167: This is not enough - the `yarn_home_dir` might exist but the config file does not. Change the code such that is checks if `yarn_site_xml_file` exists before reading, and resorts to the `YARN_CONF_DIR_ENV_VAR` if missing.

LiorMosk: Why checking if exists? it happens inside `read_config_file`, We also not check if the file exists few lines above

granulate_utils/metrics/yarn/utils.py
```diff
@@ -159,7 +161,13 @@ def detect_yarn_config(*, logger: Union[logging.Logger, logging.LoggerAdapter])
     if yarn_home_dir := find_yarn_home_dir(logger=logger):
         logger.debug(f"found YARN home dir: {yarn_home_dir}")
         yarn_site_xml_file = Path(yarn_home_dir).joinpath(RELATIVE_YARN_SITE_XML_PATH)
-        return read_config_file(yarn_site_xml_file, logger=logger)
+        if (conf := read_config_file(yarn_site_xml_file, logger=logger)) is not None:
```
Jongy, line 164: `read_config_file` API is not meant for such usage - it will log an error in this flow. I'm okay with removing the error log and adding an error log in this function (`detect_yarn_config`), something like "failed to detect config".

Jongy (changes requested):
final comment

PR #221, opened 2023-12-26
granulate-utils: Fix resolve_proc_root_links

netaneld122:
Fix resolve_proc_root_links handling of links that resolve to sub-path links.

Example:
```
{tmpdir}/link
    link -> {tmpdir}/a/c
    a -> {tmpdir}/b
Eventually we expect {tmpdir}/link to resolve to {tmpdir}/b/c
```

### DIFF
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -54,8 +54,9 @@ def resolve_host_root_links(ns_path: str) -> str:
 
 
 def abs_path_name_parts(path: str) -> Collection[str]:
-    assert os.path.isabs(path), f"Expected {path!r} to be absolute"
-    return Path(path).parts[1:]  # skip the / (or multiple /// as .parts gives them)
+    parts = Path(path).parts
+    assert parts[0].startswith("/"), f"Expected {path!r} to be absolute"
+    return parts[1:]
 
 
 def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
--- a/granulate_utils/linux/ns.py
+++ b/granulate_utils/linux/ns.py
@@ -7,9 +7,10 @@ import ctypes
 import enum
 import os
 import re
+from collections import deque
 from pathlib import Path
 from threading import Thread
-from typing import Callable, List, Optional, TypeVar, Union
+from typing import Callable, Collection, List, Optional, TypeVar, Union
 
 from psutil import NoSuchProcess, Process, process_iter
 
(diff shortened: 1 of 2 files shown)


Jongy (commented):
Didn't review yet, but just one note - there are several tests in gprofiler which test this function's behavior. Please create a PR in gprofiler which points to this commit so the tests run agains this commit. (Plus, post merging we'll want to update gprofiler as well).

granulate_utils/linux/ns.py
```diff
@@ -52,6 +53,11 @@ def resolve_host_root_links(ns_path: str) -> str:
     return resolve_proc_root_links(HOST_ROOT_PREFIX, ns_path)
 
 
+def abs_path_name_parts(path: str) -> Collection[str]:
+    assert os.path.isabs(path), f"Expected {path!r} to be absolute"
+    return Path(path).parts[1:]  # skip the / (or multiple /// as .parts gives them)
```
Jongy, line 58:
```suggestion
def abs_path_name_parts(path: str) -> Collection[str]:
    parts = Path(path).parts)
    assert parts[0] == "/", f"Expected {path!r} to be absolute"
    return parts[1:]
```
clearer?

tests/granulate_utils/test_ns_utils.py
```diff
@@ -0,0 +1,72 @@
+import os
+from pathlib import Path
+
+import pytest
+from pytest import TempdirFactory
+
+from granulate_utils.linux.ns import resolve_proc_root_links
+
+
+# Here for comparison purposes
```
Jongy, line 10: Comment "taken from revision ...."

Jongy (approved):
The core change is clear and looks good to me. I gave 2 nitpick comments. @michelhe as this is a very core part, please review as well.

granulate_utils/linux/ns.py
netaneld122: Although `isabs` deals with the case in the comment as well `//a/b` where `parts[0] == '//'` so I'll change to `startswith`

michelhe (changes requested):
Looks good, 1 small comment and a question - 
Wouldn't nsentring to the process mount namespace and then simply `readlink()` inside the namespace will serve the same purpose, but we'll have the OS do the resolving for us?

granulate_utils/linux/ns.py
```diff
@@ -62,25 +69,32 @@ def resolve_proc_root_links(proc_root: str, ns_path: str) -> str:
...
+            continue
+
+        # detect symlink loops
+        if path in seen:
+            raise RuntimeError("Symlink loop from %r" % os.path.join(path, part))
+        seen.add(path)
+
+        # resolve the part
+        ns_link = os.readlink(path)
+        if os.path.isabs(ns_link):
```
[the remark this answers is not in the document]
netaneld122: Only using pathlib for the `parts` calculation, could move it to a function so it would look more consistent, but it's all os.path.

michelhe (approved):

PR #223, opened 2024-01-16
Clear psutil's process exe cache

grariel:
Fix a bug that resulted in psutil's cache. The fix is clearing the cache before using the exe.

### DIFF
--- a/granulate_utils/linux/process.py
+++ b/granulate_utils/linux/process.py
@@ -29,7 +29,7 @@ def process_exe(process: psutil.Process) -> str:
 
     See https://github.com/giampaolo/psutil/pull/2062
     """
-    # Clear the "exe" cache on the process object. It can change after cache if it was cached during fork-exec.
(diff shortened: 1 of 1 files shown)


granulate_utils/linux/process.py
```diff
@@ -29,6 +29,8 @@ def process_exe(process: psutil.Process) -> str:
 
     See https://github.com/giampaolo/psutil/pull/2062
     """
+    # Clear the "exe" cache on the process object. It can change after cache if it was cached during fork-exec.
```
Jongy, line 32:
```suggestion
    # Clear the "exe" cache on the process object. It can change after being cached if the process execed.
```

Jongy (changes requested):
I'm fine with the change, small comment fifx

Jongy (approved):
Assuming this is sanity-tested to achieve the desired purpose

mirikl (approved):
@grariel Can you add a description to the PR?

PR #225, opened 2024-01-28
Remove backoff-stubs

michelhe:
backoff isn't required by this package, and this dev-requirement fails my CI for projects that have both `granulate-utils` and `backoff>2`


Jongy:
If CI passes 🤷

PR #229, opened 2024-02-28
Change license to Apache 2.0

motisaroka:
Change license to Apache 2.0
Also add copyright comment to each file

PR #230, opened 2024-02-28
Change license to Apache 2

motisaroka:
License to be Apache 2


Jongy:
See https://github.com/Granulate/granulate-utils/pull/229

PR #240, opened 2024-05-13
cgroups: systemd controller manipulation

glipner:
Add code to manipulate the Systemd cgroups hierarchy:.=
On Cgroups V1, manipulate the `name=systemd` controller.
On Cgroups V2, manipulate the cgroup hierarchy regardless of supported controllers (empty controller).

This is needed because Systemd keeps track of process assignment to units by their cgroups: if the V2 hierarchy is mounted, it acts as the source of truth - regardless of which controllers are enabled or delegated. If the V2 hierarchy is not mounted, a `name=systemd` controller is created to the V1 hierarchy and become the source of truth.

### DIFF
--- a/granulate_utils/linux/cgroups_v2/cgroup.py
+++ b/granulate_utils/linux/cgroups_v2/cgroup.py
@@ -28,7 +28,7 @@ from granulate_utils.linux import ns
 from granulate_utils.linux.mountinfo import iter_mountinfo
 from granulate_utils.linux.process import read_proc_file
 
-ControllerType = Literal["memory", "cpu", "cpuacct"]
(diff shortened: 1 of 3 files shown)


granulate_utils/linux/cgroups_v2/cgroup.py
```diff
@@ -225,9 +226,15 @@ def filesystem_type(self) -> str:
 
 class CgroupCoreV2(CgroupCore):
     def is_controller_supported(self, controller: ControllerType):
+        # Empty controller means we're interested just in the hierarchy
+        if controller == "":
+            return True
```
michelhe, line 231: Hmm, could have made `self.CONTROLLER` into Optional but meh.

PR #242, opened 2024-05-29
feature/containers-add-running-filter-parameter

odelame:


granulate_utils/containers/client.py
```diff
@@ -46,14 +46,29 @@ def __init__(self) -> None:
         if self._docker_client is None and self._cri_client is None:
             raise NoContainerRuntimesError()
 
-    def list_containers(self, all_info: bool = False) -> List[Container]:
+    def list_containers(self, all_info: bool = False, running: bool = False) -> List[Container]:
```
Jongy, line 49: I think that `only_running` is a better name for this parameter (which has the meaning of "If True, only running containers are returned.").

Jongy: Same for the respective docker/cri functions

Jongy (changes requested):
.

granulate_utils/containers/cri.py
```diff
@@ -63,10 +63,19 @@ def _reconstruct_name(container) -> str:
...
+                state=self.api.api_pb2.ContainerStateValue(state=self.api.api_pb2.CONTAINER_RUNNING)
+            )
+        else:
+            container_filter = None
+
         containers = []
         with self.stub() as stub:
-            for runtime_container in stub.ListContainers(self.api.api_pb2.ListContainersRequest()).containers:
+            for runtime_container in stub.ListContainers(
+                self.api.api_pb2.ListContainersRequest(filter=container_filter)
```
Jongy, line 77: is `filter` None by default? (so it makes sense to pass `None` filter?)

Jongy (approved):
lgtm, just one question left

Jongy:
I'll merge

PR #243, opened 2024-05-30
bugfix/granulate utils filter filters typo

odelame:


granulate_utils/containers/docker.py
```diff
@@ -36,13 +36,13 @@ def __init__(self) -> None:
...
-        container_filter = None
+        container_filters = None
         if only_running:
-            container_filter = {"status": "running"}
+            container_filters = {"status": "running"}
 
         containers = self._docker.containers.list(
             ignore_removed=True,
-            filter=container_filter,
+            filters=container_filters,
```
Jongy, line 45: Why didn't mypy catch it?

granulate_utils/containers/docker.py
```diff
@@ -36,13 +36,13 @@ def __init__(self) -> None:
         self._docker = docker.DockerClient(base_url="unix://" + ns.resolve_host_root_links(DOCKER_SOCK))
 
     def list_containers(self, all_info: bool, only_running: bool = True) -> List[Container]:
-        container_filter = None
+        container_filters = None
```
Jongy, line 39: From the previous PR actually, but is it None by default? Did you verify if the function accepts None vs empty dict?

Jongy (approved):
one question

PR #247, opened 2024-06-11
Cgroup2: remove bogus assumptions about cgroup mounts

glipner:
Cgroup2 detection used to look for a single mount point, that had been mounted as `mount -t cgroup2 *cgroup2* <path>` (without the `*`s).
In particular, there were two erroneous assumptions made: that the object being mount (in between `*` signs) is always called `cgroup2`, and that there might only be one cgroup2 mount present.
We've encountered case where neither of these assumptions held.

This commit removes them.


michelhe (approved):
LGTM

PR #261, opened 2024-09-18
add gprofiler utils

granulatedekel:


granulate_utils/gprofiler/utils/__init__.py
```diff
@@ -247,6 +244,7 @@ def _kill_and_reap_process(process: Popen, kill_signal: signal.Signals) -> Tuple
 
 def run_process(
     cmd: Union[str, List[str]],
+    logger: Union[logging.LoggerAdapter, logging.Logger],
```
roi-granulate, line 247: is there a reason we don’t want to enforce this as keyword argument?

Jongy:
I'm changing the source branch to `master` so we can see the total diff.

PR #262, opened 2024-09-23
databricks: Add is_databricks() based on /databricks/spark/conf/metrics.properties

Jongy:


granulate_utils/metadata/bigdata/databricks.py
```diff
@@ -15,6 +18,13 @@ def get_databricks_version() -> Optional[str]:
         return None
 
 
+def is_databricks() -> bool:
+    """
+    In some Databricks versions / images, /databricks/DBR_VERSION is missing but this file exists.
+    """
+    return os.path.exists(DATABRICKS_METRICS_PROP_PATH)
```
guybortnikov, line 25: Please use one of these https://docs.databricks.com/en/init-scripts/environment-variables.html
This metric prop isn't good enough, I don't know how reliable it is

guybortnikov (changes requested):
Tests?

guybortnikov (approved):

PR #263, opened 2024-09-23
add unedited gprofiler files in

granulatedekel:


granulate_utils/gprofiler/utils/__init__.py
```diff
@@ -0,0 +1,557 @@
...
+        ret = None
+        for cb in callbacks:
+            ret = cb()
+
+        return ret
+
+    return wrapper
+
+
+def start_process(
```
roi-granulate, line 129: a reminder to update before merging

roi-granulate:
@granulatedekel We should refrain from any mention of `gprofiler` (or any specific product) in any utility module.
while I have no problem with this PR having such `granulate_utils/gprofiler` hierarchy as a provisional step, one of your PR’s (#261 or another, new one) should take care of that.

granulate_utils/gprofiler/exceptions.py
```diff
@@ -0,0 +1,109 @@
...
+class SystemProfilerStartFailure(Exception):
+    pass
+
+
+class NoProfilersEnabledError(Exception):
+    pass
+
+
+class NoRwExecDirectoryFoundError(Exception):
+    pass
```
roi-granulate, line 109: I'll comment here because there's no changes in this file in the other PR.
note two things:
1. iiuc, we don't need all the exceptions here (we won't utilize `PerfNoSupportedEvent` for example, or am I wrong?)
2. we have `exceptions.py` in `granulate-utils` [here](https://github.com/Granulate/granulate-utils/blob/1cb5b4a7c089413679a9fdf046baa5c7ab459677/granulate_utils/exceptions.py), we can copy those exceptions to there. of course - this should be taken care of in a different PR.

roi-granulate (changes requested):
.

PR #266, opened 2025-02-04
Updated dependency versions with outstanding CSVs

awherr:
When his repo was recently moved from granulate to intel in GitHub. In the processes, IPASS scans were enabled and several CVEs in dependency versions were identified. This PR upgrades the packages with the versions that mitigate these CVEs.


mlim19 (approved):
Looks good to me. Can you explain how you figured out the versions fixing the CVEs?

